PR #8, opened 2015-04-18
Fix OCaml < 4.02.0 support

avsm:
This also adds Travis CI.  Since this library is used in some core Mirage libs now, it would be good if it could be moved to the `mirage/` org and Travis activated there.


CHANGES
```diff
@@ -1,5 +1,10 @@
+0.3.1 (2015-04-18)
```
avsm: This is wrong... looks like you hadn't updated CHANGES for the 0.3.1 release...

samoht:
I've merged that PR manually

PR #14, opened 2015-06-06
Json api

leowzukw:
It would be great to be able to write script using the number of tests passed, time spent and so on.

I started to add a json api, providing the results of the tests. But I can't call this from the command line.
I would like to do
``` sh
alcotest --api
```
to get the json output, usable by a computer, like this:
``` json
{"s":2,"f":5,"t":0.5}
```

How would I get this?

Thanks


lib/tags
```diff
@@ -0,0 +1,75 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
samoht: can you remove that file from the commit?

lib/alcotest.ml
```diff
@@ -314,6 +315,53 @@ let select_speed labels test_fun =
   if compare_speed_level (speed_of_path labels) !speed_level >= 0 then test_fun
   else skip_fun
 
+(* Type for the json api *)
+type api_content =
+    {
```
samoht: can you run `ocp-indent -i lib/alcotest.ml` to normalise the indentation?

samoht:
Thanks for your patch, that seems an interesting addition,

You need to update https://github.com/mirage/alcotest/blob/master/lib/alcotest.ml#L405 to pass the new `api` parameter around. I can have a look later this week if you didn't fix it in the meantime.

lib/alcotest.ml
leowzukw: Yes, shure.

leowzukw:
I tried several places for my `api` flag, but it doesn't build. An example with commit 4bda326.

leowzukw:
At current point, the `--api` flag prevent some messages to be displayed and outputs some JSON. But they are still some messages I can't remove. For example, the output of ```ocamlrun simple.byte --api``` (with simple.byte compiled from the example file `simple.ml`), I get : 

``` bash
$ ocamlrun simple.byte --api
test_set  0   Capitalize.                                                  [OK]
test_set  1   Add entries.                                                 [OK]
{"s":2,"f":0,"t":0.006000}
```

When I would prefer
``` bash
$ ocamlrun simple.byte --api
{"s":2,"f":0,"t":0.006000}
```

samoht:
Maybe just have a separate `print` function with a verbosity that you can control using a global variable.

samoht:
Merging now, I'll fix manually later.

leowzukw:
Ok thanks

PR #27, opened 2015-07-22
Alcotest.check: flush formatter

edwintorok:
otherwise the error message isn't actually displayed (on OCaml 4.02.2)

Previously a test I wrote failed with this rather uniformative output:
```
--------------------------------------------------------------------------------------------------------------------------------------------------------------
ASSERT same file kind
--------------------------------------------------------------------------------------------------------------------------------------------------------------

[failure] 
```

Now it fails with:
```
--------------------------------------------------------------------------------------------------------------------------------------------------------------
ASSERT same file kind
--------------------------------------------------------------------------------------------------------------------------------------------------------------

[failure] Error same file kind: expecting regular file, got directory.
```


samoht:
Thanks!

PR #28, opened 2015-07-22
Prepare the release

samoht:


samoht:
@dbuenzli any hope to see `uuseg` work on 4.00?

dbuenzli:
Don't know, I can't remember why exactly it's 4.01. But are you sure you want to make it depend on `uuseg` ? This will pull in `uucp` which doesn't use module aliases yet (because `ocamldoc` doesn't produce good documentation for them) and hence make will produce quite thick  executables.

samoht:
I just want to compute the length of an utf string.

dbuenzli:
I know, but you need data from the unicode character database for this.

samoht:
@dbuenzli ok, so after few days of very hard thinking (almost), I think I'm ok to have bigger executable if this allows users to embed emojis in the test documentation strings. Any way to port uuseg to older version of OCaml?

dbuenzli:
> if this allows users to embed emojis in the test documentation strings.

That seems like an essential feature. Did you try to rearrange your layout to avoid having to compute string lengths on non US-ASCII code points ? 

> Any way to port uuseg to older version of OCaml?

Give it a try... I have no idea, I usually dice the `ocaml-version` in my packages. I don't think there should be any GADT in there, but there are record punnings for sure.

dbuenzli:
> Any way to port uuseg to older version of OCaml?

Also note `uuseg` being dependend on `uucp`, it will in the future rather be restricted to newer versions of OCaml as I want at a certain point to switch to module aliases with `uucp` to cut down on executable sizes.

PR #29, opened 2015-07-24
Result file names

dsheets:
This patchset fixes a couple result file naming issues that have been bothering me.

First, 999a935 zero-pads the numbers in the result file names so they don't have spaces in their names which is ugly/annoying (more typing in shell for escape characters).

Then, 32cfc7b introduces best-effort alias symlink creation from `[Suite].[doc].output` to `[Suite].[00n].output`. Colliding/old files are first removed. if symlink creation fails, alcotest fails. If two tests have the same doc string, the last one to write wins. If the doc string contains bad characters, those are carried into the alias file name.

This feature is useful when each of your tests has a directory or collection of files used during the test and the doc string is reused to identify the test and to identify the resources the test uses. If it contains human-readable documentation, this is less useful.

Due to the ambiguous meaning of the doc string, I'm not sure if this design is the correct one. I certainly find it helpful but maybe the interface needs to be expanded slightly for general use. Feel free to cherry-pick the zero-padding and leave the alias generation, of course.

### DIFF
--- a/lib/alcotest.ml
+++ b/lib/alcotest.ml
@@ -23,7 +23,7 @@ type speed_level = [`Quick | `Slow]
 
 type run = unit -> unit
 
-type path = Path of (string * int)
+type path = Path of (string * int * string)
 
 type run_result = [
   | `Ok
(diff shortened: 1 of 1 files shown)


dsheets:
I also considered putting all the tests in the same suite in a subdirectory instead of using file name prefixes but I didn't do that because I don't know how useful it really is. What do you think?

samoht:
sorry I've completely miss this PR. Can you rebase it?

samoht:
> I also considered putting all the tests in the same suite in a subdirectory instead of using file name prefixes but I didn't do that because I don't know how useful it really is. What do you think?

I think that's a good idea.

samoht:
About the doc string. I'm not sure that's a so good idea to put it into the filename by default. Maybe add an option somewhere if that's useful to you.

dsheets:
How do users normally associate the test case *name* with the test case *number*?

I've never understood the enforcement of test numbers. Could each test have suite, speed, name, and doc string?

samoht:
> How do users normally associate the test case name with the test case number?

I usually don't use the log files directly, I integrate only with the command-line tool. When there are multiple errors, I just re-run the test for the given test-case only. I agree this is maybe not optimal, and I understand your PR better now :p so yes, you could keep the documentation in the file names by default.

samoht:
Can you rebase this PR? Not high-priority but after some thoughts it seems like a nice addition.

samoht:
Closing this PR as it doesn't apply anymore. Could make sense to revisit it later.

PR #33, opened 2015-09-16
add boolean assert

zjhmale:
maybe there should be a boolean assert to check boolean values


zjhmale:
@samoht

samoht:
Thanks! There are few others combinators which might be useful here:

https://github.com/samoht/ciso/blob/master/lib_test/test_common.ml#L18-L41


I'll add them shortly.

zjhmale:
cool!

samoht:
alcotest 0.4.5 is now released in opam with this patch.

PR #36, opened 2015-10-31
add missing newline to verbose output

seliopou:
The first dashed line for a test case starts on the same line as the start event. This adds the newline to fix that.


samoht:
Thanks! Feel free to add more combinators as well :-)

PR #37, opened 2015-11-10
result: add Result.result combinator

seliopou:


seliopou:
Any interest in this patch? I'll be releasing a library soon that uses alcotest and depends on this. If it's not going to get merged, I can just move the code into the tests for that library.

samoht:
Sorry, I've completely missed your PRs...

PR #39, opened 2015-12-26
Fix output file contents

dsheets:
The current output file contents are mostly useless in the case of `Alcotest`-generated error messages. Lots of colorful successful assertion messages are printed but the actual cause of test failure is omitted.

This patch causes error and exception messages to be printed to the output file when the output is active. It also uses a single open fd to the result file for both stdout and stderr in order to have the message streams co-operate regarding file offset.


samoht:
Thanks!

PR #41, opened 2016-01-16
Fix typo in -o option specification

aantron:
Without this fix, the `-o` option looks like:

    ---o=DIR (absent=./_tests/)

with it:

    -o DIR (absent=./_tests/)


dsheets:
Thanks!

PR #43, opened 2016-01-25
Change list TESTABLE

ruhatch:
Changed list TESTABLE to use the equality of the underlying element, which is necessary for cases when (=) does not work (e.g. Cstruct list)


dsheets:
Nice! This is much better. Thank you!

PR #44, opened 2016-01-26
use Astring instead of Stringext (only String.cut is used here, hardly..

hannesm:
a reason to depend on some other library anyways)


hannesm:
but this means to no longer work with OCaml 4.00... can likely be ignored, this PR..

samoht:
Alcotest now uses Astring.

PR #50, opened 2016-03-14
pass: add a testable that always passes

seliopou:
Use pass to disregard components of data types you are testing. For example, if you want to test that a pair-valued function, but disregard its second component for the in the test, you can use the following:

```ocaml
Alcotest.(pair int pass)
```


samoht:
Thanks!

PR #51, opened 2016-03-18
Fix index out of bounds from empty test doc string

dariusf:
This happens if we supply an empty string as the first element of the test case tuple.


samoht:
Thanks!

Issue #52, opened 2016-04-01
Add a color option and/or respect TERM variable.


Drup:
Jenkins doesn't interpret the escape sequences, it's an unreadable mess :(

PR #55, opened 2016-04-23
Fixes

samoht:


samoht:
I have also patches to fix #52 (but not on the right computer...), would push and make a new release on Monday.

PR #57, opened 2016-05-01
support 4.3: -warn-error is enabled (which I like), fix deprecated warnings

hannesm:
Char.uppercase is deprecated, use Char.uppercase_ascii insteead
String.uppercase is deprecated, use String.uppercase_ascii insteead

an alternative would be to disable warn-error for warning 3 (deprecation)


Drup:
I think it's more reasonable to make warning 3 non-fatal. Seding into ocaml code in a makefile gives me all kind of chills.

hannesm:
@Drup see #58 whatever people (well, @samoht) prefers..

hannesm:
I prefer #58... @samoht

PR #58, opened 2016-05-01
make warning 3 (deprecation) non-fatal

hannesm:


hannesm:
I could with 4.3... travis complained at https://travis-ci.org/hannesm/ocaml-hkdf/jobs/126948371

PR #59, opened 2016-05-03
Prepare the release

samoht:


hannesm:
... but this will lead to 4.03 only (since Char/String.uppercase_ascii was introduced in 4.03 and is not present in earlier releases) ...

samoht:
Indeed, sorry I think I need more coffee.

hannesm:
hmm, and indeed my intuition with semantics of command-line arguments were wrong, thanks for spotting this :+1:

PR #63, opened 2016-05-24
Add an optionnal argument `arg`

dinosaure:
It's useful when you try to use `Cmdliner`in top of `Alcotest`.

### DIFF
--- a/lib/alcotest.mli
+++ b/lib/alcotest.mli
@@ -35,7 +35,9 @@ val run: ?and_exit:bool -> ?argv:string array -> string -> test list -> unit
     happens when the function ends. By default, [and_exit] is set,
     which makes the function exit with [0] if everything is fine or
     [1] if there is an issue. If [and_exit] is [false], then the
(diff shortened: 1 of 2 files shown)


lib/alcotest.mli
```diff
@@ -29,7 +29,7 @@ type test = string * test_case list
 exception Test_error
 (** The exception return by {!run} in case of errors. *)
 
-val run: ?and_exit:bool -> string -> test list -> unit
+val run: ?and_exit:bool -> ?argv:string array -> string -> test list -> unit
```
mor1: Also good to update the doc string ?

talex5:
Minor typos:

The optional argument [argv] speci**fies** the argument**s** sen**t** to alcotest like ["--json"], ["--verbose"], etc. (**at least one argument is required**)

(also, **optional** in the commit message)

Apart from that, LGTM.

samoht:
Thanks! I'll fix the docs in a separate commit.

PR #75, opened 2016-10-24
Simpler

pqwy:
It's much easier to add types with a `testable` constructor and two projections. This adds them.

It also simplifies definitions by using them.

And it simplifies some bits further by fully committing to `Fmt`, which is a dep anyway.


talex5:
LGTM! And much cleaner :-)

samoht:
Thanks!

PR #77, opened 2016-10-24
Detect _build and move output there

### LINKED ISSUE #76: Would be nice to place output under `_build` if present
`_build` is automatically cleaned by `ocamlbuild`. If `_tests` were under `_build`, when the latter is present, there would be one less temporary directory to take care of.

pqwy:
Checks if there is `_build` in the current dir. Dumps outputs in `_build/_tests` if so. Fixes #76.

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -154,8 +154,20 @@ let file_of_path path ext =
 
 let output_file t path = Filename.concat t.test_dir (file_of_path path "output")
 
+let mkdir_p path mode =
+  let rec mk parent = function
+  | [] -> ()
+  | name::names ->
+      let path = parent ^ "/" ^ name in
+      begin try if not (Sys.is_directory path) then
+        Fmt.strf "mkdir: %s: is a file" path |> failwith
+      with Sys_error _ -> Unix.mkdir path mode end;
+      mk path names in
+  match String.cuts ~empty:true ~sep:"/" path with
+  | ""::xs -> mk "/" xs | xs -> mk "." xs
+
 let prepare t =
-  if not (Sys.file_exists t.test_dir) then Unix.mkdir t.test_dir 0o755
+  if not (Sys.file_exists t.test_dir) then mkdir_p t.test_dir 0o755
 
 let color c ppf fmt = Fmt.(styled c string) ppf fmt
 let red_s fmt = color `Red fmt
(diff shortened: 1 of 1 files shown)


avsm:
dislike changing the defaults on the basis of the build directory existing like this. why not explicitly specify the test output directory in the pkg.ml?

pqwy:
Because it needs to be specified in 1-15 places per project, for every project. It's just enough noise that it's too much.

Agree with magic detection, but can we do _anything_ here not do default to `alcotest` claiming its own dir? Unconditionally go in `_build`?

pqwy:
To clarify, the problem I have is that adding `alcotest` means that I have to tweak `.gitignore`, and I get a build-related directory that is not maintained, or cleaned, by the build system. In routine development the contents of this directory is of no interest. So I found myself constantly deleting it.

Compare `OUnit`:
```
    List.find
      dir_exists
      [
        concat pwd "_build";
        concat (dirname pwd) "_build";
        concat (dirname (dirname pwd)) "_build";
        pwd
      ]
```

samoht:
> Unconditionally go in _build?

Yes, I think that makes sense to me. We can change the default from `_tests` to `_build/_tests`.

pqwy:
This?

In the unlikely case that someone pulled out the test executable and deleted `_build`, or didn't use `ocamlbuild` in the first place, it dies like:

```
test.native: internal error, uncaught exception:
             Unix.Unix_error(Unix.ENOENT, "mkdir", "_build/_tests")
```

... but that's perfectly fine by me.

pqwy:
Can I do something to help move this ahead? Add recursive dir creation if `_build` does not exist? Do something else altogether?

samoht:
Sorry, I missed your comment. Yes, recursive `mkdir` seems a good idea to me.

pqwy:
Ok -- mkdir is currently `Unix.mkdir`. Without extra deps, there's two ways to get a recursive one:

* call out to `mkdir -p`
* split the path on `/` and `Unix.mkdir` all prefixes.

Not sure how the first one plays with Windows, but `-p` is pretty standard in `mkdir` implementations. The second one is robust, but only if the Windows folks use forward slashes.

Any preference?

pqwy:
Here's a recursive mkdir to recursively mkdir `_build/_tests`!

pqwy:
@samoht ping

samoht:
Thanks! And sorry again for the delay.

samoht:
Released in https://github.com/ocaml/opam-repository/pull/7739

pqwy:
Thanks!

PR #78, opened 2016-10-25
reduce the use of polymorphic comparison internally

avsm:
There should be no semantic change here, just more explicit use
of comparison functions


avsm:
Oh never mind, this breaks on <4.03.0, so probably not worth it for now.

PR #81, opened 2016-10-25
Add Rresult_msg testable

avsm:


samoht:
I am not totally convinced by add a new dependency to `rresult` but maybe that's fine. Let me think a bit more about it.

samoht:
I am not very convinced by that PR :-) It seems to me that:

```
let rresult_msg = testable Rresult.R.pp_msg (=)
```

Is easy enough to add in your tests. I'm happy to reconsider when `rresult` is used more widely and that everyone is force to duplicate that line in all of their tests.

PR #83, opened 2016-11-09
Clean up handling of env variables

samoht:


samoht:
Not sure if I have done the right thing, see dbuenzli/cmldiner#64

PR #89, opened 2017-04-17
Allowed passing absolute error to Alcotest.float

superbobry:
Due to approximate nature of floats direct equality is rarely useful in practice. This PR adds a basic form approximate equality: "equal up to an absolute error".

There is one more step we can take to make testing floating point code with Alcotest a breeze: generalize `Alcotest.float` to support support relative error and "equal up to [ULP](](https://en.wikipedia.org/wiki/Unit_in_the_last_place))" via polymorphic variants, e.g.

```ocaml
Alcotest.float (`Ulp 1)
Alcotest.float (`Abs 1)
Alcotest.float (`Rel 1)
```

What do you think?

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -543,7 +543,7 @@ let int64 = testable Fmt.int64 (=)
 
 let int = testable Fmt.int (=)
 
-let float = testable Fmt.float (=)
+let float eps = testable Fmt.float (fun x y -> abs_float (x -. y) < eps)
 
 let char = testable Fmt.char (=)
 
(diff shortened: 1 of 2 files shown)


samoht:
Sorry for the delay, but that looks like a pretty good idea! Thanks for the patch.

superbobry:
Thanks @samoht! That he patch only implements absolute difference. Do you think I should add an extended version described in the above comment?

samoht:
I don't write libraries using floats, so I don't know what people would expect :-) I thing having an epsilon makes sense but if you feel that someone more structured is needed, please submit a patch.

PR #96, opened 2017-06-20
Allow the test-case to take a state parameter as argument

### LINKED ISSUE #93: allow command-line arguments to be passed to tests?
We have [tests](https://github.com/mirage/mirage-tcpip/blob/master/lib_test/test_iperf.ml#L197) which might generate some artifacts useful for correcting the program behavior that caused the test to fail.  It would be great to be able to pass command-line arguments to the test code specifying whether these artifacts should be generated, and if so, where to store them, but currently we have no way to do this for our Alcotest tests.

samoht:
This state parameter is built by looking at the CLI. Use `Alcotest.run_with_args`
to use this, as `Alcotest.run` will continue to pass `()` as test parameters.

Fix #93

### DIFF
--- a/lwt/alcotest_lwt.ml
+++ b/lwt/alcotest_lwt.ml
@@ -14,7 +14,7 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-let run fn () =
+let run fn args =
   let async_ex, async_waker = Lwt.wait () in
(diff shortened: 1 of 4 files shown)


samoht:
/cc @yomimono 

This is a bit more intrusive as I would like as it change the base types from `(unit -> unit)` into `('a -> unit)`. It's a good change but that will maybe break some libraries. I will try to make a revdep test to see the breakages...

Issue #101, opened 2017-07-09
Documentation Down


kennetpostigo:
It seem the documentation is down. Has it by any chance been moved elsewhere?

samoht:
There were indeed a small glitch, should be fixed now! Thanks for reporting the issue and sorry for the delay.

PR #102, opened 2017-07-17
Fix broken link to documentation

smeruelo:
This should solve #100 and #101


samoht:
Thanks a lot!!

smeruelo:
Oops, bad URL.. sorry, and thanks for fixing it :)

PR #104, opened 2017-07-24
Async support in the style Alcotest_lwt

rgrinberg:


async/alcotest_async.mli
```diff
@@ -0,0 +1,7 @@
+
+val test_case
```
samoht: Can you add some documentation?

rgrinberg: Sure, added the copyright header as well

samoht:
Thanks!

Issue #97, opened 2017-06-20
When running from jbuilder, Alcotest shows old results


talex5:
I've noticed that when I do e.g.

```
jbuilder build --dev test/test.bc
./_build/default/test/test.bc test core
```

Alcotest often displays two errors. It looks like one is from the run I just did, and the second is from a previous run. Any idea what could cause that?

talex5:
Actually, maybe this is a different error. Let me investigate a bit further...

talex5:
Sorry, today's problem was something different. Will reopen if I can reproduce the error I thought I saw before.

talex5:
OK, I can reproduce it now - not imagining things!

```
tal@laptop /p/t/alco> cat test.ml 
let tests = [
  "Hello", `Quick, (fun () -> Alcotest.fail "BLUE BLUE BLUE");
]

let () =
  Alcotest.run "test" [
    "tests", tests;
  ]
```

```
tal@laptop /p/t/alco> cat jbuild 
(jbuild_version 1)

(executable (
  (name test)
  (libraries (alcotest))
))
```

```
tal@laptop /p/t/alco> jbuilder build --dev test.exe; and ./_build/default/test.exe -ev
------------------------------------------------------------------------------------------------------------------------------
ASSERT BLUE BLUE BLUE
------------------------------------------------------------------------------------------------------------------------------
Testing test.
 ...                tests          0   Hello.
[ERROR]             tests          0   Hello.
-- tests.000 Failed --
Hello.
_build/_tests/tests.000.output:
-----------------------------------------------------------------------------------------
ASSERT RED RED RED
-----------------------------------------------------------------------------------------
Test error: Error RED RED RED.

Test error: Error BLUE BLUE BLUE.

The full test results are available in `_build/_tests`.
1 error! in 0.000s. 1 test run.
```

```
tal@laptop /p/t/alco> cat _build/_tests/*
-----------------------------------------------------------------------------------------
ASSERT RED RED RED
-----------------------------------------------------------------------------------------
Test error: Error RED RED RED.
```

`RED RED RED` does not exist anywhere in the code. That text only existed in a previous version of the source. Pretty confusing when it's pages of log output...

samoht:
It seems that `-v` doesn't create the test logs anymore. I'm investigating to find why ...

samoht:
[...]

PR #108, opened 2017-08-03
Do not display outdated logs when using `-v`

samoht:
The redirect logic is still very hairy, but this looks like the minimal amount of changes needed to clarify that `-v` doesn't save the log results to a file, and thus should not display the log results from that (very possibly outdated) file.

Should address #97

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -200,23 +200,23 @@ let print_info t p =
 let left_c = 20
 
 let error t path fmt =
-  let filename = output_file t path in
-  let output =
-    if not (Sys.file_exists filename) then "--"
-    else
-      let file = open_in filename in
-      let output = string_of_channel file in
-      close_in file;
-      output
-  in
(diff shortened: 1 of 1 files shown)


src/alcotest.ml
```diff
@@ -375,13 +374,16 @@ let show_result t result =
...
     in
-    Fmt.(pf stdout) "The full test results are available in `%s`.\n\
-                     %a in %.3fs. %d test%s run.\n%!"
-      t.test_dir msg () result.time result.success (s result.success)
+    let full_logs =
+      if t.verbose then ""
+      else Fmt.strf "The full test results are available in `%s`.\n" t.test_dir
+    in
+    Fmt.(pf stdout) "%s%a in %.3fs. %d test%s run.\n%!"
+      full_logs test_results () result.time result.success (s result.success)
```
talex5: You can use `%t` for a pretty-printer with no argument.

PR #133, opened 2018-05-20
Improve documentation on test case filtering and flush error formatter on exit

edwintorok:
Filtering testcases was scarcely documented (had to read the source code to find out the expected format): add an example to the README, and make --help more explicit that up to 2 arguments can be used for filtering (the name of the test, and the test case index).

Also fix a bug with error messages getting lost on exit: OCaml doesn't flush Format.stderr on exit, just Format.stdout, so we have to flush it ourselves.

The improved output compared to the old one looks like this:
```diff
--- old
+++ new
@@ -1,15 +1,19 @@
 + ./simple test test_set invalidnumber
+simple: TESTCASE argument: invalid value `invalidnumber', expected an integer
+Usage: simple test [OPTION]... [NAME] [TESTCASE]
+Try `simple test --help' or `simple --help' for more information.
 Testing My first test.
 + echo '$?=' 1
 $?= 1

 + ./simple test nonexistent_name
+Invalid request (no tests to run, filter skipped everything)!
 Testing My first test.
 + echo '$?=' 1
 $?= 1

 + ./simple test test_set 1 trailingarguments
-simple: internal error, uncaught exception:
-        Failure("filter_test")
-
+simple: too many arguments, don't know what to do with `trailingarguments'
+Usage: simple test [OPTION]... [NAME] [TESTCASE]
+Try `simple test --help' or `simple --help' for more information.
 Testing My first test.
 + echo '$?=' 1
 $?= 1
```

No doubt the documentation could be improved further, this is just a start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,9 @@ $ ./simple.native
 Test Successful in 0.001s. 2 tests run.
 ```
 
-You can filter which tests to run by exact name and/or test case number:
+You can filter which tests to run by supplying either the exact test name
+(which would run all testcases with that name), or the exact test name
+and test case number (which would run just that single test):
 ```shell
 $ ./simple.native test test_set
 Testing My first test.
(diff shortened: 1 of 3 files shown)


samoht:
Thanks!

PR #134, opened 2018-05-28
Add a (partial) comparison between testing frameworks

### LINKED ISSUE #132: Missing comparison with other testing frameworks
It would be great if the README (or another good place in the documentation) could mention a short comparison to other common testing frameworks, such ounit. What are the advantages? What are the drawbacks? When to use alcotest, when to use ounit?

samoht:
Fix #132 

/cc @vog

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,3 +177,34 @@ in _build/_tests/all.000.output:
 freeing all resources
 [failure] All is broken
 ```
+
+### Comparison with other testing frameworks
+
+The README is pretty clear about that:
+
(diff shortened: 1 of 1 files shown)


vog:
Looks great!

I found two minor typos:

1)

> Alcotest is the only testing fraweworks using colors!

Correction:

> Alcotest is the only testing framework using colors!

2)

> qcheck does random generator

Correction:

> qcheck does random generation

samoht:
Thanks, typo fixed!

Issue #135, opened 2018-09-10
tests location hint is wrong with dune


samoht:
See https://github.com/samoht/dune-release/pull/69#issuecomment-419905777

m-harrison:
The problem here is that `dune` appears to alter the current working directory as part of `dune build @runtest`. When `default_cmd` calls into `run_registred_tests` the `Cmdliner` parameter of `test_dir` gets evaluated. `dune` doesn't provide a `-o` parameter and so the default value is used. This default value is a relative path and so this relative path is going to be traversed from the current working directory set by `dune`.

The result of this is that as observed, the test logs get dropped in `/Users/matt/git/{project name}/_build/default/lib_tests/_build/_tests`, but when reporting the test location back out to the user the current code reports the literal `t.test_dir` which is the relative path - hence the confusing user experience.

The fix I'm proposing (and that I'll add to my existing PR) is to rather than default `test_dir` to the relative path `"_build/_tests"`, make sure that it gets set to the absolute path prefixed by the current working directory.

Issue #125, opened 2018-04-18
Create a fresh log sub-dir for every run


samoht:
This would be useful to compare various run. 

The sub-dir should be a timestamp or a uuid, and it should be displayed at the beginning of the run (so that even if an user press CTRL-C the location of the logs is still available to read)

PR #137, opened 2018-10-13
Creates a fresh log sub-directory on each run

m-harrison:
#125 
--
This proposed change takes a dependency on `uuidm` to generate UUIDs. It extends the `alcotest` global state to hold a globally unique run ID. This ID is newly generated on each call into `run_with_args` and is displayed out to the user before running any tests. The output logs from the tests are then placed in this sub directory which itself is under the `-o` directory (or under the default test directory if this is not specified).

To test this change I pinned my local clone with the changes, built the test executable of another project and repeatedly manually invoked it. The output logs were subsequently generated in separate sub-directories corresponding to the ID of the run, as expected.

#135 
--
This was caused by `dune` altering the current working directory before running tests. It would not explicitly pass in a `-o` parameter and so the default one was taken which was a relative path. The result of this was that a relative path was printed to the user and while this was correct it wasn't obvious that it was from a different current working directory. My fix made the default parameter an absolute path under the current working directory with the same relative path (functionally this is a no-op change). This meant when printed to the user it is now correct.

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -65,6 +65,7 @@ type 'a t = {
   json       : bool;
   verbose    : bool;
   test_dir   : string;
+  run_id     : Uuidm.t;
 
 }
 
@@ -81,9 +82,10 @@ let empty () =
   let show_errors = false in
   let json = false in
   let test_dir = Sys.getcwd () in
+  let run_id = Uuidm.nil in
   { name; errors; tests; paths; doc; speed;
     max_label; speed_level;
-    show_errors; json; verbose; test_dir }
+    show_errors; json; verbose; test_dir; run_id }
 
 let compare_speed_level s1 s2 =
   match s1, s2 with
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -65,7 +65,7 @@ type 'a t = {
   json       : bool;
   verbose    : bool;
   test_dir   : string;
-  run_id     : Uuidm.t;
+  run_id     : string;
 
 }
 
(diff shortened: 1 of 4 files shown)


src/alcotest.ml
```diff
@@ -383,7 +390,9 @@ let show_result t result =
     in
     let full_logs ppf =
       if t.verbose then Fmt.string ppf ""
-      else Fmt.pf ppf "The full test results are available in `%s`.\n" t.test_dir
+      else
+        Fmt.pf ppf "The full test results are available in `%s`.\n"
+          (output_dir t)
```
samoht: Not directly related to that PR but #135 points that this location is in general wrong. Might be a good occasion to fix that too, but I am fine if this is done in a different PR too!

m-harrison: Good call out, see the comment I left on #135 for my analysis of the issue. The summary is that I believe `dune` alters the current working directory and as the default `test_dir` is a relative rather than an absolute path it appears to be wrong (the problem being that in the context of the test runner it is correct).

My fix (pushed in `fbfb395`) is to set this default `test_dir` to an absolute path whose prefix is the current working directory. This means semantically there is no change but when `t.test_dir` is printed to the user it is explicit about the current working directory (and just the entire location). It also means the current working directory is taken once at the start of the run rather than implicitly picking it up every time that `t.test_dir` was referenced in some I/O.

m-harrison:
> That looks great, thanks a lot for your patches! I've just pushed a fix for the CI, hopefully everything should be green now.

Thanks for the review! It looks like the CI is still red so I'll leave that with you. I don't have write access to the repo so will need you to merge anyway. If you want to keep this one open a while to try and get the CI working then feel free!

samoht:
Thanks a lot for your patch! I'll draft a new release with your changes.

Issue #140, opened 2018-11-02
output dir broken on Windows/Cygwin


kkirstein:
On Windows/Cygwin (mingw-toolchain) I get an error when running tests with the default output dir:
```
test_runner.exe: internal error, uncaught exception:
                 Unix.Unix_error(12, "mkdir", "./C:\\Users\\<..>\\_build\\default\\test\\_build")

Testing Benchmark.
This run has ID `A376A064-4F4C-4A56-91A7-0DFEE54FAEB7`.
```
It looks like there is a mix-up between relative and absolute paths together with the (usual) Windows path separator `\` issue. `test_runner --help` reveals also a broken path as default for `-o`. Calling `test_runner` manually with `-o` and a correct path works as expected.

Since 0.8.3 is working correctly, I suspect that 32690c6e279b85eb8a9b129550acef73b10855c8  has introduced this regression (most probably by calling `Sys.getcwd`). But instead of fixing there, it might be better to make `mkdir_p` function in `alcotest.ml` platform dependent.

kkirstein:
Just tested the merged master on Windows 10 (using Cygwin-bash, cmd.exe & WSL-bash) -> ok.

PR #141, opened 2018-11-02
fix output path for windows/cygwin

kkirstein:
This is a fix for #140:

- hard-coded path separator "/" has been replaced by `Filename.dir_sep`
- `mkdir_p` now can deal with Windows absolute paths, containing drive letters, e.g., C:\

Tested on Windows 10 cygwin/mingw toolchain & Windows 10 WSL aka Bash for Windows (Ubuntu 18.04)

Default path shown by `test_runner --help` still looks broken, but this is only a cosmetic issue (probably needs more escaping for "\\")

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -163,16 +163,20 @@ let output_file t path =
   Filename.concat (output_dir t) (file_of_path path "output")
 
 let mkdir_p path mode =
+  let sep = Filename.dir_sep in
   let rec mk parent = function
   | [] -> ()
   | name::names ->
-      let path = parent ^ "/" ^ name in
+      let path = parent ^ sep ^ name in
       begin try if not (Sys.is_directory path) then
         Fmt.strf "mkdir: %s: is a file" path |> failwith
       with Sys_error _ -> Unix.mkdir path mode end;
       mk path names in
-  match String.cuts ~empty:true ~sep:"/" path with
-  | ""::xs -> mk "/" xs | xs -> mk "." xs
+  match String.cuts ~empty:true ~sep:sep path with
+  | ""::xs -> mk sep xs
+  (* check for Windows drive letter *)
+  | dl::xs when Str.string_match (Str.regexp "[A-Z]:") dl 0 -> mk dl xs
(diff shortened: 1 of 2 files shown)


src/alcotest.ml
```diff
@@ -163,16 +163,20 @@ let output_file t path =
...
       begin try if not (Sys.is_directory path) then
         Fmt.strf "mkdir: %s: is a file" path |> failwith
       with Sys_error _ -> Unix.mkdir path mode end;
       mk path names in
-  match String.cuts ~empty:true ~sep:"/" path with
-  | ""::xs -> mk "/" xs | xs -> mk "." xs
+  match String.cuts ~empty:true ~sep:sep path with
+  | ""::xs -> mk sep xs
+  (* check for Windows drive letter *)
+  | dl::xs when Str.string_match (Str.regexp "[A-Z]:") dl 0 -> mk dl xs
```
samoht: would it be possible to not depend on `Str` for this? It adds a dependency to a C binding (`-lcamlstr`) and this is not thread-safe. Maybe https://github.com/dbuenzli/fpath/blob/master/src/fpath.ml#L41 is enough here.

samoht:
Many thanks for your contribution! I've added a comment about the use of `Str`.

Once this is merged (I guess pretty soon :p) I'll add an appveyor CI to be sure this is no regression on windows.

src/alcotest.ml
samoht: Sorry I wasn't very clear: you don't need to bring the full dependency to fpath here; just copy the same logic here should be sufficient and would avoid growing the dependency too much.

kkirstein: Yes, I like the idea of keeping the number of deps small (I could have noticed that at the first time...).
At the end, it will be a rewrite of `mkdir_p` function. Might take another day or two (got busy with another task).

kkirstein: OK, this should now be better: No additional dependency, just relying on the `astring `package (which is already a dep). 358a4750a06389526b53f6bd93520e0c5313e6ca

samoht:
Thanks!

Issue #146, opened 2019-02-22
A more quiet output


emillon:
Hi,

When using alcotest, I noticed that it is very verbose when there are no failures. When there are many tests and they are run in a tight loop, it's not immediately obvious if there are failures or not.

Ideally, what I'm after is a mode where only failures are displayed (or something very simple like one dot per passing test case).

I'm happy to provide a PR to add a flag if this is something you agree with.

Thanks!

samoht:
Having a "quiet" mode makes sense to me (for instance using `--mode quiet` or `--quiet` if that doesn't conflict with the current flags); 

I kind of like when I see lots of green checkboxes ("e.g. lots of things work great!") but I understand there are use-cases when you just wants to see nothing when everything is good.

CraigFe:
Closing as #149 added a `--silent` flag.

PR #149, opened 2019-04-25
Add silent option

andersfugmann:
This PR adds option `--silent` which will suppress output for all successful tests.
It contains two commits - one which removes an unused global mutable variable (quiet), and one to add option `--silent`
This addresses #146. 

I did not want this PR to be too intrusive, so there are some `if t.silent then ...` scattered around the code.  Also I would have preferred to not have the first two lines printed (see examples below), but these are printed before any command line arguments are parsed, and I did not want to change the current flow.

Example when all tests are successful:
```
$ dune exec drivers/yaml/test/unittest.exe -- --silent
Testing yaml.
This run has ID `E09E988F-081E-4894-A7CB-E2D3FCFB60A5`.
...............................................................
```

Example with one failing test:
```
$ dune exec drivers/yaml/test/unittest.exe -- --silent
Testing yaml.      
This run has ID `835BF394-F1E5-476D-8A51-14C558AC4D4D`.
................
[ERROR]             Maybe fail                      0   Fail test case.
.............................................
-- Maybe fail.000 [Fail test case.] Failed --
in /home/afu/git/ppx_protocol_conv/_build/_tests/835BF394-F1E5-476D-8A51-14C558AC4D4D/Maybe fail.000.output:
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
ASSERT Blah
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
[failure] Error Blah: expecting
0, got
1.
Raised at file "stdlib.ml", line 33, characters 17-33
Called from file "src/alcotest.ml", line 290, characters 8-14


The full test results are available in `/home/afu/git/ppx_protocol_conv/_build/_tests/835BF394-F1E5-476D-8A51-14C558AC4D4D`.
1 error! in 0.005s. 62 tests run.
```

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -259,10 +259,11 @@ let print_event t = function
     print_info t p;
   | `Result (_, r) when t.silent -> begin
       let c = match r with
-        | `Ok -> "."
-        | `Error _ | `Exn _ -> "E"
-        | `Skip -> "S"
-        | `Todo _ -> "T"
+        | `Exn _   -> "F"
+        | `Error _ -> "E"
+        | `Skip    -> "S"
+        | `Todo _  -> "T"
+        | `Ok      -> "."
       in
       print t (fun ppf -> Fmt.string ppf c)
     end
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -62,10 +62,9 @@ type 'a t = {
   show_errors: bool;
   json       : bool;
   verbose    : bool;
-  silent     : bool;
+  compact    : bool;
   test_dir   : string;
   run_id     : string;
-
 }
 
 let empty () =
@@ -77,7 +76,7 @@ let empty () =
   let tests = [] in
   let max_label = 0 in
   let verbose = false in
-  let silent = false in
+  let compact = false in
   let speed_level = `Slow in
   let show_errors = false in
   let json = false in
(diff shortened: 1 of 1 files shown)


emillon:
From a quick read this looks like what I was after in #146, thanks!

@nathanreb should be interested in reviewing this 😃

andersfugmann:
I like the printing of 'E' for errors. I will update the PR to do that.

andersfugmann:
I've updated the PR to reflect your comments. I also solved a bug introduced with the change to print 'E' for errors, as I did not notice that the function `error` had the side-effect of storing the stringified error. I've changed it to only have this side effect, and print status lined explicitly. 

I've also renamed the option to be `--compact`

src/alcotest.ml
```diff
@@ -227,38 +226,51 @@ let error t path fmt =
...
     print_info t p
   | `Todo _        ->
     print t (fun ppf -> left left_c yellow_s ppf "[TODO]");
     print_info t p
 
+let print_compact_result t = function
+  | `Exn _   -> print_ch t "F"
+  | `Error _ -> print_ch t "E"
+  | `Skip    -> print_ch t "S"
+  | `Todo _  -> print_ch t "T"
```
NathanReb: Awesome ordering!

PR #152, opened 2019-06-14
Add checks for infinity and NaN in float check

psafont:
Until now some comparisons between edge-cases in floats were failing.

A test suite for floats has been added in `examples/` to easily compare the before and after.

Edit: Reflecting back on this, there might be reason to keep the standard behaviour, I'm open to suggestion on how to best keep both behaviours.

### DIFF
--- a/examples/dune
+++ b/examples/dune
@@ -1,3 +1,3 @@
 (tests
- (names simple)
+ (names simple floats)
  (libraries alcotest))
--- a/examples/floats.ml
+++ b/examples/floats.ml
@@ -1,0 +1,62 @@
+(*
+This is free and unencumbered software released into the public domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
+
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
+software to the public domain. We make this dedication for the benefit
+of the public at large and to the detriment of our heirs and
+successors. We intend this dedication to be an overt act of
+relinquishment in perpetuity of all present and future rights to this
+software under copyright law.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+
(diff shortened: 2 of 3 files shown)


examples/dune
```diff
@@ -1,7 +1,7 @@
 (executables
-  (names     bad simple)
+  (names     bad simple floats)
   (libraries alcotest))
 
 (alias
   (name runtest)
-  (deps bad.exe simple.exe))
+  (deps bad.exe simple.exe floats.exe))
```
emillon: That's not really related to your changes, but I don't think that this runs the examples. Can you confirm that introducing a runtime error in the test is not caught by `dune runtest`?

examples/floats.ml
```diff
@@ -0,0 +1,64 @@
...
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+
+For more information, please refer to <http://unlicense.org/>
+*)
+
+let e = epsilon_float
+
+(* The tests *)
```
emillon:
```suggestion
```

examples/floats.ml
```diff
@@ -0,0 +1,64 @@
...
+let edge_set = [
+  "NaN", `Quick, nan     ;
+  "∞",   `Quick, infinity;
+]
+
+let others_set = [
+  "others", `Quick, others;
+]
+
+(* Run it *)
```
emillon:
```suggestion
```

src/alcotest.ml
```diff
@@ -617,7 +619,12 @@ let int64 = testable Fmt.int64 (=)
 
 let int = testable Fmt.int (=)
 
-let float eps = testable Fmt.float (fun x y -> abs_float (x -. y) <= eps)
+let float eps =
+  let same x y = (isnan x && isnan y) ||
```
emillon: what do you think about naming each case to help reading operators? Something like:
```ocaml
let both_are_nan = isnan x && isnan y in
...
both_are_nan || ...
```

examples/dune
psafont: None of the examples are run on `dune runtest` as there is no action specified for runtest dune file in there.
This line could be added to run simple and floats: `(action (progn (run ./simple.exe) (run ./floats.exe)))`

src/alcotest.ml
psafont: Does it bring that much clarity? I'm not convinced

examples/dune
emillon: I'll fix that separately (the idiomatic way to do this is to turn the executable stanza into a test stanza)

src/alcotest.ml
emillon: I think that it would help with understanding operator priority and add an opportunity to explain why the `x = y` case is important.

psafont: Ideally I would use `Float.equal x y || abs_float (x -. y) <= eps`.

In any case I'm not sure how would using `both_are_nan` would clarify the second term is related to infinity values, will a comment do?

emillon: Sure, `Float.equal` works for me!

psafont: Sorry, I forgot to say why that's the ideal solution: it's not available on Ocaml < 4.07. (4.04 to 4.06 seem to be supported)

samoht:
Thanks for your patch, that sounds like a good idea! 

Would you be able to rebase this to master now that #153 is merged?

samoht:
Thanks!

PR #153, opened 2019-06-18
Run some examples in `runtest`

emillon:
Hi!

This splits the `examples/` directory in two: the "good" set is registered as tests, so that it is executed by `dune runtest`. The "bad" example is just built by `dune runtest`.

(see #152)

Thanks!


samoht:
Thanks!

PR #155, opened 2019-07-14
Test output dirs: Create symlinks for each test suite + "latest"

### LINKED ISSUE #147: test output directory is confusing
since #125 / #137 was merged, the test output is located at `_build/default/test/_build/_tests/UUID/<testname.txt>`. This is very hard to navigate to.

In order to read the output of a failing test, I now first have to figure out the UUID of this run, which is sometimes hidden pages above the failing test case. And then I've to navigate to the directory, remembering on the next test run to switch to a different directory again.

I still don't understand the motivation to output into a randomly-generated directory, but if it is useful for others, what about creating a symbolic link when a testrun starts(!!) linking to the current testrun e.g. `_build/default/test/current` pointing to `_build/_tests/UUID/`? then there'd at least be a canonical way to read the output!?

cfcs:
Test output directories: Create symlinks for each test suite, and a "latest" symlink.

- The "latest" symlink points to the last test executed. This is useful to quickly navigate to whichever directory contains failing tests.

- Additionally a symlink named after t.name (in turn named after argv[0] of the test executable) is created. This is useful when there are multiple test executables being compiled as part of a build chain.

- Also try to avoid creating world-writable directories and files by setting umask and changing explicit permissions to og-rwx. This is best-effort, but since OCaml doesn't provide adequate bindings to the Unix system calls it is hard to perform operations like deleting files and creating directories recursively without introducing TOCTOU race conditions that can be exploited by less-privileged processes running on the machine.

- After executing my test suite three times, my test dir now looks like:
```shell
$ ls /home/user/ocaml/xi-rope/_build/_tests/ -la
total 28
drwx------ 5 user user 4096 Jul 14 05:48 .
drwxr-xr-x 7 user user 4096 Jul 14 05:47 ..
drwx------ 2 user user 4096 Jul 14 05:48 A3DC6B25-317A-4DA8-8BF8-FA2FE0FD5A3E
drwx------ 2 user user 4096 Jul 14 05:48 CDF189EE-35C2-43E8-A6A0-D62F86BED1B3
drwx------ 2 user user 4096 Jul 14 05:48 F4100A63-F959-47E5-889A-311F29BB2594
lrwxrwxrwx 1 user user   75 Jul 14 05:48 latest -> /home/user/ocaml/xi-rope/_build/_tests/F4100A63-F959-47E5-889A-311F29BB2594
lrwxrwxrwx 1 user user   75 Jul 14 05:48 tests.native -> /home/user/ocaml/xi-rope/_build/_tests/F4100A63-F959-47E5-889A-311F29BB2594
```

fixes #147

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -189,7 +189,19 @@ let mkdir_p path mode =
 
 let prepare t =
   let test_dir = output_dir t in
-  if not (Sys.file_exists test_dir) then mkdir_p test_dir 0o770
+  if not (Sys.file_exists test_dir) then begin
+    mkdir_p test_dir 0o770 ;
+    if Sys.unix || Sys.cygwin then begin
+      let this_exe = Filename.concat t.test_dir t.name
+      and latest = Filename.concat t.test_dir "latest" in
+      if Sys.file_exists this_exe then Sys.remove this_exe ;
+      if Sys.file_exists latest then Sys.remove latest ;
+      Unix.symlink ~to_dir:true test_dir this_exe ;
+      Unix.symlink ~to_dir:true test_dir latest ;
+    end
+  end else if not (Sys.is_directory test_dir) then
+    failwith (Fmt.strf "exists but is not a directory: %S" test_dir)
+
 
 let color c ppf fmt = Fmt.(styled c string) ppf fmt
 let red_s fmt = color `Red fmt
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -189,7 +189,7 @@ let mkdir_p path mode =
 
 let prepare t =
   let test_dir = output_dir t in
-  if not (Sys.file_exists test_dir) then mkdir_p test_dir 0o755
+  if not (Sys.file_exists test_dir) then mkdir_p test_dir 0o770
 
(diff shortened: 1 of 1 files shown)


hannesm:
the final output looks good to me, I'm curious about the permission changes of the output files (setting umask / 0600 instead of 0666) -- they imho deserve their own commit

cfcs:
There's an overlap of lines which would lead to a merge conflict, and in any case I don't think it makes sense to introduce more unsafe file system handling functions before changing umask etc.

cfcs:
Ah, first though @hannesm wanted two PRs; but after talking to him split into two commits instead. :)

talex5:
Why override the user's umask? Alcotest only creates world-writeable directories if you have explicitly set your umask to request that, and if you've configured it like that by mistake then exposing your unit-test results is unlikely to be your biggest problem...

cfcs:
@talex5 that is a valid point, I think I got something confused between the world-writable files `0o666` and the world-readable directories `0o755` that it was creating. I've removed the `umask` call and force-pushed. Thanks for noticing :)

hannesm:
to me this looks good! would make my life easier every day I use alcotest and try to find the testruns.

samoht:
Thanks for the patch!

Issue #156, opened 2019-07-25
Support for running multiple names / testcases via the CLI


CraigFe:
Currently the `test` CLI allows for picking a particular test name to run, and optionally a specific test case number within that test. There is no support for running multiple test names, or multiple case numbers within a given test case number. Some suggested changes:

 - Support multiple test case numbers as a comma-separated list, potentially including ranges (e.g. `test.exe-test FOO 1,6-10,15`).
 - Support multiple test names as a comma-separated list (e.g. `test.exe-test FOO,BAR`). This would be a breaking change for projects with commas in their test names already; however, I suspect that's fairly rare.

These features would be useful for testing interactions between subsets of a large test suite, as in [mirage/irmin](https://github.com/mirage/irmin/). Happy to do it myself, but interested in others' opinions first. 

CC: @samoht.

samoht:
The first item sounds good. For the second item I'm not sure -- a regex for test names would probably make more sense.

CraigFe:
Hmm... yes, agreed that it makes more sense. It also conflicts with a greater range of existing test names. In particular, `.` becomes the regex wildcard. I know that we're using that in Irmin as a pseudo-namespacing thing. We'd probably want to make the regex parsing optional and off by default?

samoht:
I'm not sure it'll break anything: `GIT.HTTP` will only match with one type of tests for instance.

cfcs:
1) That sounds like an excellent suggestion!
2) 1) What do you propose doing in cases like `test.exe test    A,B,C    1,3-5`?
    2) It seems to me this can easily be handled in the shell too:
        ```shell
         eval './_build/default/lzo/tests.exe test '{AFL,Hardcoded}\;
        ```

    2) I think using a regex sounds like a good idea. As brought up it will "break" backwards compatibility where people are using `.` to run specific tests named like `a.c` and at the same time have a test case called like `abc`. It will "break" by running `abc` in addition to `a.c`, but `a.c` will still get run. I think that seems like an acceptable change in behavior, and it is easy to adapt to in whatever tooling might hit this edge case.

     3) IIRC I had some trouble somewhere using test names with slashes (`/`) in them due to the test name being used literally in paths. Maybe we could kill two birds with one stone by using `/` as separator (instead of comma), and make slash illegal in test names. In any case maybe this warrants an issue of its own...


[...]

PR #157, opened 2019-07-25
Add support for executing subsets of tests via CLI

CraigFe:
This is a candidate solution for #156.

CC: @samoht

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,27 +24,36 @@ A simple example:
 
 (* A module with functions to test *)
 module To_test = struct
-  let capit letter = Char.uppercase letter
-  let plus int_list = List.fold_left (fun a b -> a + b) 0 int_list
+  let lowercase = String.lowercase_ascii
+  let capitalize = String.capitalize_ascii
+  let str_concat = String.concat ""
+  let list_concat = List.append
 end
 
 (* The tests *)
-let capit () =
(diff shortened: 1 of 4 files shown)


CraigFe:
This behaviour probably warrants some unit testing.

Happy to do that in either this PR or another. Since there are currently no unit tests at all, it didn't feel like it belonged here.

samoht:
That looks great! Can you add some documentation and a few tests?

CraigFe:
Having thought a bit about setting up unit testing for Alcotest, I'm not sure how best to do it. It feels wrong to test Alcotest.dev with Alcotest.dev, but I don't know of another way to do set things up in Dune. Perhaps we should use `ppx_expect` instead?

samoht:
Not totally sure about testing -- for now, just run a few manual tests I'd say. But most importantly, do document these new features in the README and in the man pages.

samoht:
Can you rebase this one? I'm not totally sure about the new dependency to `re` but I think it's worth it.

CraigFe:
Rebased, and also changed the example in the README. It's now more verbose (I think it was actually so concise as to not compile before!), but demonstrates a non-trivial use of regular expressions.

Let me know what you think! :slightly_smiling_face:

samoht:
That looks great! Thanks

PR #158, opened 2019-07-25
dune build @fmt: upgrade to ocamlformat.0.10

CraigFe:
I suggest the use of ocamlformat 0.10, now that it has been released.

Big QOL improvement at the cost of a big diff.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,2 @@
+version = 0.10
+profile = conventional
--- a/async/alcotest_async.ml
+++ b/async/alcotest_async.ml
@@ -6,11 +6,10 @@ let run timeout name fn args =
   Thread_safe.block_on_async_exn (fun () ->
       Clock.with_timeout timeout (fn args) >>| function
       | `Result x -> x
-      | `Timeout -> Alcotest.fail (
-          Printf.sprintf "%s timed out after %s"
-            name (Time.Span.to_string_hum timeout)
-        )
-    )
+      | `Timeout ->
+          Alcotest.fail
+            (Printf.sprintf "%s timed out after %s" name
+               (Time.Span.to_string_hum timeout)))
 
-let test_case ?(timeout=sec 2.) name s f =
+let test_case ?(timeout = sec 2.) name s f =
   Alcotest.test_case name s (run timeout name f)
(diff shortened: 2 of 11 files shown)


cfcs:
Big diff, and lost ability to use `git blame` when analyzing regressions.
I think something like `ocamlformat` sounds nice for individual patches, but doing a large pass like this kind of ruins a lot of the benefits you get from using version control.
I'm a bit skeptical that is worth the QOL (what does QOL mean?)

Arguably `git blame` is not the cleverest tool in the toolbox, doing line-by-line comparisons.
Do you know if there's a way to make `git blame` work despite reformatted code, ie at higher granularity than line-by-line?

CraigFe:
QOL = quality of life. Currently, to work with this code I have to either not use a formatter or manually rein it in so that it doesn't touch extra lines. I agree that it only trades one tooling problem for another (adding noise to the `git log`) but it has the advantage of being an O(n) benefit for an O(1) upfront cost. I personally think the benefits outweigh the costs (particularly for a pre-1.0 library), but opinions may differ :slightly_smiling_face: If the consensus is for keeping the history pure, I'm more than happy to drop this PR. There is also the nuclear option of `git filter-branch` to rewrite history so that each commit is prettified, but that brings the usual concerns with rebasing.

Regarding `git blame`: personally I use `git blame <format_rev>~ <path>` to backtrack past reformat commits whenever they're getting in the way. As far as I'm aware, `git` has no semantic blaming tool (although that would be cool!).

cfcs:
1) I guess it will always lead to more merge conflicts though, since each patch for `n` lines will now be touching `n+reformatted` lines?
2) Ah, of course @ `rev~` / `rev~99`~ It still seems like a bit more work, but I guess that is okay. And for `alcotest` I guess this is in any case not so relevant since we don't have to deal with backwards compatibility for file formats.

Anyway, it's not my library and I'm not a regular contributor, so I'm not the person to ask whether it makes sense for `alcotest`, I just saw this scrolling past in the notifications tab and left my initial thoughts here :)  Thanks for the clarifications!

samoht:
LGTM

CraigFe:
Rebased

Issue #160, opened 2019-07-30
Success exit code on failing to register tests


CraigFe:
Currently, registering a test may fail if the test's name contains non-ascii characters. This results in an error message being printed and the remaining tests being execute, but does not result in a failing exit code. 

```
[ craigfe@carbon ] ~/t/alcotest                                                                                                                 [3570054]
ᐅ dune exec test/main.exe
Error: "\239\129\173" is not a valid test label (it should be an ASCII string), skipping.
Testing alcotest.
This run has ID `F38E390E-2877-4F31-B6F7-6A23B68E001E`.
[OK]                bar1          0   unix.
[OK]                foo1          0   lwt.
The full test results are available in `/home/craigfe/t/alcotest/_build/_tests/F38E390E-2877-4F31-B6F7-6A23B68E001E`.
Test Successful in 0.000s. 2 tests run.

[ craigfe@carbon ] ~/t/alcotest                                                                                                                 [3570054]
ᐅ echo $?
0
```

In my opinion, this is a bug since various CI tools use this code to greenlight a commit.

CraigFe:
Fix now merged.

Issue #115, opened 2017-09-12
Test case names are used directly for filesystem interactions


cfcs:
Hello! I'm taking Alcotest for a spin, and my very first test case happened to be en-/decoding of values.

So I wrote up:
```ocaml
let test_int () =
  Alcotest.(check int) "same int" 1234 (1234)

let tests =
  [
    "integer en/decoding",
      ["check that ints are good", `Slow, test_int ];
  ]

let () =
  Alcotest.run "my test suite" tests
```

and it gave me this error:
```
$ topkg te
Testing my test suite.
 ...                integer en/decoding          0   check that ints are good.alcotest_lib.native: internal error, uncaught exception:
                     Unix.Unix_error(Unix.ENOENT, "open", "_build/_tests/integer en/decoding.000.output")
                     
pkg.ml: [ERROR] Some tests failed.
```

This turned out to be caused by the slash in the test case name. Changing it to `integer en- and decoding` made it work. It was a little bit counter-intuitive to me as a new user of the library, but I guess it's a corner case - wanted to let you know either way. I like all the colors! :-)

CraigFe:
Thanks! This has been fixed by a whitelist of characters in test names, and a line added in the README, so hopefully this will be less counter-intuitive in future :slightly_smiling_face:

PR #161, opened 2019-07-30
Handle filenames with slashes correctly

CraigFe:
This is a candidate solution for #160 and #115.

If any of the tests fail registration, Alcotest will search for any additional registration errors, print them to stderr, and then exit with code 1. Alcotest now also gives an error if a test name contains forward or backward slashes (as these names must be valid folder names -- see #115).

Original behaviour:
```
[ craigfe@carbon ] ~/t/alcotest
ᐅ dune exec test/main.exe
Error: "\239\129\173" is not a valid test label (it should be an ASCII string), skipping.
Testing alcotest.
This run has ID `CC560E17-DF73-485F-A026-EF15BAEBB9B1`.
[OK]                bar1                0   unix.
[OK]                foo1                0   lwt.
 ...                with/slash          0   name_with_slash/foo.main.exe: internal error, uncaught exception:
          Unix.Unix_error(Unix.ENOENT, "open", "/home/craigfe/t/alcotest/_build/_tests/CC560E17-DF73-485F-A026-EF15BAEBB9B1/with/slash.000.output")
```

New behaviour:
```
[ craigfe@carbon ] ~/t/alcotest
ᐅ dune exec test/main.exe
Error: "\239\129\173" is not a valid test label (it should be an ASCII string).
Error: "with/slash" is not a valid test label (contains an illegal character).

[ craigfe@carbon ] ~/t/alcotest
ᐅ echo $?
1
```

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -601,7 +601,7 @@ let int_range_list =
     let acc i = (set := IntSet.add i (!set)) in
     let ranges = String.cuts ~sep:"," s in
     let process_range s =
-      let bounds = String.cuts ~sep:"-" s |> List.map (String.to_int) in
+      let bounds = String.cuts ~sep:".." s |> List.map (String.to_int) in
       match bounds with
       | [Some i] -> acc i
       | [Some lower; Some upper] when lower <= upper ->
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -476,16 +476,19 @@ let list_tests t () =
 
 let is_ascii s = String.for_all Char.Ascii.is_valid s
 
-let err_ascii s =
-  let err =
-    Printf.sprintf
-      "%S is not a valid test label (it should be an ASCII string), skipping." s
-  in
-  Fmt.(pf stderr) "%a %s\n%!" red "Error:" err
+let validate_name name =
+  let errmsg reason = Fmt.strf "%a %S is not a valid test label (%s)." red "Error:" name reason in
+  if not (is_ascii name) then
+    Error (errmsg "it should be an ASCII string")
+  else
+    Ok ()
 
 let register t name (ts: 'a test_case list) =
-  if not (is_ascii name) then (err_ascii name; t)
-  else (
+  match (t, validate_name name) with
+  | Error error_acc, Error e -> Error (e::error_acc)
+  | Error error_acc, Ok ()   -> Error error_acc
+  | Ok _, Error e            -> Error [e]
+  | Ok t, Ok () ->
     let max_label = max t.max_label (String.length name) in
     let paths = Hashtbl.create 16 in
     let docs = Hashtbl.create 16 in
@@ -505,8 +508,7 @@ let register t name (ts: 'a test_case list) =
     let paths = t.paths @ paths in
     let doc p = try Some (Hashtbl.find docs p) with Not_found -> t.doc p in
     let speed p = try Some (Hashtbl.find speeds p) with Not_found -> t.speed p in
-    { t with paths; tests; doc; speed; max_label; }
-  )
+    Ok { t with paths; tests; doc; speed; max_label; }
 
 exception Test_error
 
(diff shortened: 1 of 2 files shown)


hannesm:
it is not immediately clear to me whether this is a complete solution for #115, i.e. have you thought about:
- test name being `.`
- test name being `..`
- two test names being case-insensitive equal (`foo`, `fOo`), and executing on a case-insensitive file system (IIRC macos still ship with such a thing by default)
- any other control character 0x00..0x20?

CraigFe:
It's a good concern to have; you're right that these cases are probably not being handled correctly. I had a look some sort of filename validation in [Filename](https://caml.inria.fr/pub/docs/manual-ocaml/libref/type_Filename.html), but couldn't find any. 

If possible, I'd like to avoid considering these edge cases explicitly in Alcotest. Do you have any suggestions?

hannesm:
Well, my main point was to figure out what is the problem space here. 
- A simple solution (breaking change) would be to not use test suite names as filenames, but instead numbers or hashes thereof.
- Another solution may be to have a whitelist of allowed characters?
- A third solution would be to url-percent-escape (%7F) "evil"/"bad" characters (or String.escape?).

I don't have sufficient time to figure out the design tradeoffs, which would involve to potentially break a lot of user intuition -- it may be worth to do some empirical evaluation of "bad/evil characters/strings used in test suite names of alcotest" first!?

hannesm:
(NB of course this PR is already an improvement of the current state, but I'm usually more interested in solutions that solve an issue once and for all -- excuse me for being pedantic ;)

src/alcotest.ml
```diff
@@ -463,16 +463,25 @@ let list_tests t () =
...
 let is_ascii s = String.for_all Char.Ascii.is_valid s
 
-let err_ascii s =
-  let err =
-    Printf.sprintf
-      "%S is not a valid test label (it should be an ASCII string), skipping." s
-  in
-  Fmt.(pf stderr) "%a %s\n%!" red "Error:" err
+let contains_illegal_char s =
+  let disallowed = ['/'; '\\'] in
```
samoht: I think I'd prefer to have a whitelist, e.g. change `is_ascii` to actually check that the test name contains only "valid" characters (e.g. `a-z` + `A-Z`  + `0-9` + a few chars (maybe `.` and '-`' and `_`). (and not `..`)

samoht: Actually I'm fine to disallow `.` as well. Can be fixed easily in most of my libs anyway :-)

CraigFe: Agreed. This doesn't deal with the case-insensitive filesystem problem mentioned by @hannesm (which I haven't tested, but seems plausible), but we can worry about that separately.

CraigFe: Will wait until #157 has been merged to make use of the `re` depencency.

CraigFe:
- Rebased
- Change to use a whitelist of characters instead
- Added a line in the README explaining the new limitation

Also:
- Changed the selective range syntax to `..` rather than `-` (on the suggestion of @cfcs [here](https://github.com/mirage/alcotest/issues/156#issuecomment-516049169)), to bring it in line with the README. Forgot to do this in #157... sorry!

samoht:
Thanks! I think we should also probably normalise filenames at one point, but I am not sure exactly how, and this could be done later.

samoht:
After a bit more thoughts, I am not sure breaking existing tests is a good idea -- and that people should be free to use whatever unicode characters as they want to name their tests.

So instead, we should make sure the filename are normalised properly, by replacing weird characters by something different (for instance `-`), something like:

```ocaml
let slugify s =
  let last_char_is_dash = ref false in
  let s = String.Ascii.lowercase s in
  let buf = Buffer.create (String.length s) in
  String.iter
    (function
      | ('a' .. 'z' | '0' .. '9') as c ->
        if !last_char_is_dash then (
          Buffer.add_char buf '-';
          last_char_is_dash := false );
        Buffer.add_char buf c
      | _ -> last_char_is_dash := true)
    s;
  Buffer.contents buf
```

CraigFe:
I like this idea, but it has the snag of `slugify` not being injective: what will happen if multiple test names get mapped to the same directory? (such as if someone were to use emoji for all of their test names).

samoht:
good point -- maybe just add a unique id/counter at the beginning of the filename.

CraigFe:
Sounds good to me. I suggest adding the UID at the end of the filename, so that it doesn't mess with tab-completion.

emillon:
This may be a naive question, but why do tests write to the disk?

CraigFe:
The log output of every test is written to disk separately, I assume as a debugging mechanism

Now that you mention it, this should potentially be an opt-in feature? I just checked my disk, and each Irmin `runtest` writes 8.5MB of logs, even if everything is passing.

samoht:
You need a way to keep track of logs to show them back to the user in case some of them fail -- to inspect CI issues, etc.

Issue #128, opened 2018-04-20
The time reported to run tests is wrong


samoht:
It happens at least for lwt tests:

```
$ time _build/default/foo.exe
Testing foo.
[OK]                all          0   one.
The full test results are available in `_build/_tests`.
Test Successful in 0.001s. 1 test run.

real    0m10.024s
user    0m0.004s
sys     0m0.008s
```

CraigFe:
The fix for this has now been merged. Closing.

PR #162, opened 2019-07-30
Report real time rather than system time

CraigFe:
This is a candidate solution for #128.

Currently the test length is measured in CPU time rather than real time, leading to unintuitive output when running tests that are not CPU-bound (particularly Lwt tests). This changes the internals to measure time with `Unix.time` instead.

Original behaviour:
```
let () =
  Alcotest.run "alcotest" [
      ("bar1", [ Alcotest.test_case "unix" `Quick (fun () -> Unix.sleep 1)]);
      ("foo1", [ Alcotest_lwt.test_case "lwt" `Quick (fun _ () -> Lwt_unix.sleep 1.)]);
    ]

Test Successful in 0.000s. 2 tests run.
```
New behaviour:
```
Test Successful in 2.000s. 2 tests run.
```

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -445,11 +445,11 @@ let show_result t result =
 
 let result t test args =
   prepare t;
-  let start_time = Sys.time () in
+  let start_time = Unix.time () in
   let test = map_test (redirect_test_output t) test in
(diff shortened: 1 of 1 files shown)


samoht:
Good catch. thanks!

Issue #126, opened 2018-04-18
Review how tests are registered


samoht:
Few users have expressed the fact that listing all the tests is error prone (as you could easily forget to add a test if you don't have error-as-warning + an empty mli to guide you).

Maybe we expose a combinator like crowbar where tests are automatically added to a global list of things to test.

aantron:
One potential issue here is that if tests are spread out over multiple files, either (1) the files have to still be referenced from somewhere to be linked, or (2) we need to tell users to pass `-linkall`.

yawaramin:
I'd like to suggest a design for adding to the 'automatic list of things to test'. In the below hypothetical code snippet I will use the test suite shown in the readme example: https://github.com/mirage/alcotest/blob/main/README.md#examples

```code
Alcotest.suite "Utils" begin fun group ->
  group "string-case" begin fun case ->
    case "Lower case" begin fun () ->
      Alcotest.(check string) "same string" "hello!" (To_test.lowercase "hELLO!")
    end;

    case "Capitalization" begin fun () ->
      Alcotest.(check string) "same string" "World." (To_test.capitalize "world.")
    end;
  end;

  group "string-concat" begin fun case ->
    case "String mashing" begin fun () ->
      Alcotest.(check string) "same string" "foobar" (To_test.str_concat ["foo"; "bar"])
    end;
  end;

  group "list-concat" begin fun case ->
    case ~speed:`Slow "List mashing" begin fun () ->
      Alcotest.(check (list int)) "same lists" [1; 2; 3] (To_test.list_concat [1] [2; 3])
    end;
  end;
end
```

The idea of this style is that Alcotest calls the given test suites and groups with functions which have been already set up internally with `_ list ref` and on each call just adds the relevant item to its corresponding list. Basically save users from having to mess around with lists, and instead raise the conceptual level Alcotest suites, groups, and cases.

ELLIOTTCABLE:
I quite like @yawaramin's approach; it looks (and feels) a lot more like the RSpec / Jest style tests that I've been used to over the years from other languages and environments.

(In fact, I think Jest even used to provide _both_ styles of API, with the `describe`/`it(…)`-style API being optional sugar? I think that'd be a great idea here as well.)

yawaramin:
Hey, yeah, I had something like Ruby on my mind for this. To make it easy for people to compare, pretending for a second that the module was written in Ruby, here's part of it translated to RSpec:

```ruby
RSpec.describe Utils, "Utils" do
[...]

PR #167, opened 2019-08-02
Generalise test logic on a concurrency monad

### LINKED ISSUE #119: Making Alcotest more asynchronous
`alcotest-lwt` currently [calls `Lwt_main.run`](https://github.com/mirage/alcotest/blob/ebe9cf2be006c626a6994dd4f684b8c6f722b141/lwt/alcotest_lwt.ml#L17-L29) for each test, in order to convert a Lwt test that is executed asynchronously into a synchronous one that completes before the function returns.

We would like to use `alcotest-lwt`, or similar modules, in Lwt and elsewhere, but `Lwt_main.run` is not available on Node.js or in browsers – there is no way to convert asynchronous code to synchronous by waiting on something.

We basically would need Alcotest to have fully-asynchronous tests, and to push `Lwt_main.run` to one place as far to the end of the test runner as possible. Ideally, `Alcotest.run` itself would return a promise.

Something similar had to be done to the Lwt tester recently, and it is what is preventing us from replacing it with Alcotest: https://github.com/ocsigen/lwt/commit/5a272758adbdf9bf21a874377605b224e0753851#diff-fa96e02582bba62b4c7d8bac70d3e8e7.

cc @kennetpostigo

CraigFe:
This resolves #119 by parameterising the core testing logic on a concurrency monad. This allows `Alcotest_lwt` and `Alcotest_async` to provide `run` methods that return promises, rather than using `Lwt_main.run` and `Thread_safe.block_on_async_exn` internally. As discussed in #119, these scheduling methods present problems for Javascript/Mirage deployments.

This PR is currently not `ocamlformat`-ted to minimise the diff while it's under review.

### API changes
- Standard `Alcotest` keeps the same API, with no changes.

- `Alcotest.run` can now no longer run `Alcotest_lwt` test cases, which must instead be passed to `Alcotest_lwt.run` (see [this diff](https://github.com/mirage/alcotest/compare/master...CraigFe:monad#diff-0f4ab61e93c78dfaac56b35c0c0ed0f5)). The asynchronous backends now provide an unimaginatively named `test_case'` fallback for running synchronous tests (avoiding the user boilerplate of lifting these tests inside the monad). Is there a better name for this method?

### DIFF
--- a/src/alcotest.mli
+++ b/src/alcotest.mli
@@ -29,16 +29,22 @@
 
     {e Release %%VERSION%% } *)
 
+exception Test_error
+(** The exception return by {!run} in case of errors. *)
+
 type speed_level = [ `Quick | `Slow ]
 (** Speed level of a test. Tests marked as [`Quick] are always run. Tests marked
-   as [`Slow] are skipped when the `-q` flag is passed. *)
+    as [`Slow] are skipped when the `-q` flag is passed. *)
 
-type 'a test_case = string * speed_level * ('a -> unit)
+module type TESTER = sig
+  type return
+
+type 'a test_case = string * speed_level * ('a -> return)
 (** A test case is an UTF-8 encoded documentation string, a speed
     level and a function to execute. Typically, the testing function calls the
     helper functions provided below (such as [check] and [fail]). *)
 
-val test_case : string -> speed_level -> ('a -> unit) -> 'a test_case
+val test_case : string -> speed_level -> ('a -> return) -> 'a test_case
 (** [test_case n s f] is the test case [n] running at speed [s] using
     the function [f]. *)
 
@@ -46,11 +52,8 @@ type 'a test = string * 'a test_case list
 (** A test is an US-ASCII encoded name and a list of test cases.
  * The name can be used for filtering which tests to run on the CLI *)
 
-exception Test_error
-(** The exception return by {!run} in case of errors. *)
-
 val run :
-  ?and_exit:bool -> ?argv:string array -> string -> unit test list -> unit
+  ?and_exit:bool -> ?argv:string array -> string -> unit test list -> return
 (** [run n t] runs the test suite [t]. [n] is the name of the
     tested library.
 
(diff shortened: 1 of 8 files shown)


src/alcotest.mli
```diff
@@ -74,11 +77,28 @@ val run_with_args :
...
+
+module type MONAD = sig
+  type 'a t
+
+  val return : 'a -> 'a t
+
+  val bind : 'a t -> ('a -> 'b t) -> 'b t
+end
+
+module MonadicTester (M : MONAD) : TESTER with type return = unit M.t
```
samoht: It's best to avoid camel-case (especially in the API). I guess here just call it `Make (M: MONAD)`

src/alcotest.mli
```diff
@@ -29,28 +29,31 @@
...
 
+exception Test_error
+(** The exception return by {!run} in case of errors. *)
+
 type speed_level = [ `Quick | `Slow ]
 (** Speed level of a test. Tests marked as [`Quick] are always run. Tests marked
-   as [`Slow] are skipped when the `-q` flag is passed. *)
+    as [`Slow] are skipped when the `-q` flag is passed. *)
+
+module type TESTER = sig
```
samoht: `module type S = sig` would also work here

samoht:
Thanks!

CraigFe:
Oops, this wanted an `ocamlformat` before being merged; sorry! I'll make a PR now

aantron:
I do suggest renaming `test_case'`. Perhaps `test_case_sync`? Better options welcome.

aantron:
(I want to note that in practice, however, it's not a big deal to lift test cases into the monad. It's clearer, and doesn't take that much more typing than a different function name.)

CraigFe:
I like `test_case_sync` more than `test_case'`, certainly. This scheduling interface is slated for review in https://github.com/mirage/alcotest/issues/126, which might make these questions redundant. If you have suggestions for larger changes to the registration mechanism, I suggest taking them there; I'll make a draft PR as soon as I have a sufficiently convincing idea :smile:

The nice aspect of an extra `test_case_sync` (rather than just doing the lifting manually) is that it also avoids creating the extra `Lwt.switch` and forcing synchronous test cases to ignore a switch argument.

aantron:
I don't see a discussion of a scheduling interface in #126, but this reminded me that in some libraries (like Lwt and Luv), it would be nice to be able to express which tests can run concurrently. Some tests have deliberate time delays, or long delays due to I/O, and these can add up when all the tests are run in sequence.

My vague idea was some kind of `lock` argument. By default, all tests can run concurrently (in practice, they will run in sequence if your underlying monad never yields, so "simple" users won't see a difference). Tests specifying equivalent `lock`s have to be serialized.

aantron:
On a second thought, the lock itself doesn't need Alcotest support. It rather needs the ability to be yielded to to spawn more concurrent tests, and the library under test can provide its own locking inside the body of each test case that needs it.

aantron:
I haven't thought it through, but this may mean an explicit applicative interface, or something close to it.

PR #169, opened 2019-08-07
Add CI runners for OS X and Windows

CraigFe:
It would be nice to have CI on multiple OSs, since we currently use the file system to store log outputs of tests.

Currently just experimenting with TravisCI multi-OS support; not intended for merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,19 @@
 language: c
-install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
-script: bash -ex .travis-docker.sh
-services:
-  - docker
-sudo: false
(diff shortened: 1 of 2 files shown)


CraigFe:
Both new CI runners are now passing. 

Currently the new OSs are just testing core `Alcotest` (not `Alcotest_lwt` or `Alcotest_async`). Even still, this adds 15 minutes to a successful CI run, since the OS X runners provided by Travis are about 3x slower than their Linux runners. Testing the async backends on OS X would add a further ~1 hour (since OS X runners are sequential), which I think is not worth it.

I'd appreciate a review from someone more familiar with OCaml CI. Perhaps @avsm or @samoht?

samoht:
That looks good. Slowing down the CI is a bit sad, but I hope we can fix that later.

PR #176, opened 2019-08-12
Refactor test registration logic

### LINKED ISSUE #170: Log capturing fails on case-insensitive filesystem
Following on from issues raised in #115, Alcotest behaves poorly on a case-insensitive filesystem. Consider the following example:
```
let () =
  Alcotest.run "Case-insensitive FS"
    [ ( "alpha", [ Alcotest.test_case "a" `Quick test_a ] );
      ( "ALPHA", [ Alcotest.test_case "b" `Quick test_b ] )
    ]
```
This results in only a single log being persisted (`alpha.000.output`). If an error occurs during the first tests, it is impossible to see the full trace of that error.

### LINKED ISSUE #175: Silently ignoring duplicate test names
If Alcotest is run on a suite with duplicate `(test_name, index)` cases, the log output of the last instance will overwrite those of the former. Consider the following example:

```
  Alcotest.run "Utils"
    [ ( "duplicated_name",
        [ Alcotest.test_case "Lower case" `Quick test_lowercase;
          Alcotest.test_case "Capitalization" `Quick test_capitalize
        ] );
      ( "duplicated_name",
        [ Alcotest.test_case "String mashing" `Quick test_str_concat ] )
    ]
```

This produces only two logs: `duplicated_name.000.output` and `duplicated_name.001.output`.

CraigFe:
This fixes #170 and fixes #175 by adding checks for filepath collisions whenever a test is registered.

Also refactors the test registration / caching logic to a separate `Suite` module and simplifies the logic  somewhat. This module is a candidate for splitting to a separate file, but doing so would require moving the types somewhere else too.

### DIFF
--- a/src/alcotest.ml
+++ b/src/alcotest.ml
@@ -113,15 +113,51 @@ module Make (M : MONAD) = struct
 
   type 'a test = string * 'a test_case list
 
+  module Suite : sig
+    type 'a t
+
+    val empty : unit -> 'a t
+
+    val add : 'a t -> path * string * speed_level * 'a rrun -> 'a t
+
+    val tests : 'a t -> (path * 'a rrun) list
+
+    val doc_of_path : 'a t -> path -> string
+
+    val speed_of_path : 'a t -> path -> speed_level
+  end = struct
+    type 'a t = {
+      tests : (path * 'a rrun) list;
+      (* caches computed from the library values. *)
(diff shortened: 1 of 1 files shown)


CraigFe:
cc/ @samoht. On a related note, how do you feel about splitting into multiple files?

samoht:
Looks good, thanks

PR #178, opened 2019-08-13
Initial end-to-end test suite

CraigFe:
This initialises an end-to-end test suite that compares example executions against expected outputs. The process is driven by auto-generated `dune` files, allowing a workflow of `dune runtest` →  `dune promote`. 

Currently this does not allow for passing options to the executable under test, which would be a nice addition. For this, we could use extra `*.opt` files similarly to [how it's being done in OCamlformat](https://github.com/ocaml-ppx/ocamlformat/tree/master/test/passing). 

### Frustrations
1. The `dune` action DSL has no way of negating the exit code of a command, so it's being done manually with a little script -- I might request this as a Dune feature...

1. It's necessary to sanitize the output of the tests to remove environment dependencies / randomness. Again, this is being done by an extra script.

1. I initially added symlinks to the `examples/` directory to verify their output, but these appear to break the Windows CI, so I removed them.

### DIFF
--- a/test/e2e/expect_failure.ml
+++ b/test/e2e/expect_failure.ml
@@ -1,0 +1,9 @@
+(* Run an executable and negate its return status *)
+let () =
+  let args = List.tl (Array.to_list Sys.argv) in
+  let command = Fmt.strf "%a" Fmt.(list ~sep:(const string " ") string) args in
+  let rel_command = Filename.(concat current_dir_name) command in
+  match Unix.system rel_command with
+  | WEXITED 0 -> exit 1
+  | WEXITED _ -> exit 0
+  | WSIGNALED _ | WSTOPPED _ -> failwith "Process terminated by a signal"
--- a/test/e2e/failing/dune
+++ b/test/e2e/failing/dune
@@ -1,0 +1,17 @@
+(include dune.inc)
+
+(rule
+ (targets dune.gen)
+ (deps (source_tree .))
+ (action
+  (with-stdout-to
+    %{targets}
+   (run ../gen_dune_rules.exe --expect-failure)
+  )
+ )
+)
+
+(alias
+ (name runtest)
+ (action (diff dune.inc dune.gen))
+)
--- a/test/e2e/gen_dune_rules.ml
+++ b/test/e2e/gen_dune_rules.ml
@@ -1,0 +1,87 @@
+let global_stanza filenames =
+  let bases = List.map Filename.remove_extension filenames in
+  let pp_sexp_list = Fmt.(list ~sep:(const string "\n   ")) in
+  Fmt.pr
+    {|(executables
+ (names
+   %a
+ )
+ (libraries alcotest)
+ (modules
+   %a
+ )
+)
+|}
+    (pp_sexp_list Fmt.string) bases (pp_sexp_list Fmt.string) bases
+
+let example_rule_stanza ~expect_failure filename =
+  let base = Filename.remove_extension filename in
+  let expect_failure =
+    if expect_failure then "../expect_failure.exe " else ""
+  in
+  (* Run Alcotest to get *.actual, then pass through the strip_randomness
+     sanitiser to get *.processed. *)
+  Fmt.pr
+    {|
+(rule
+ (target %s.actual)
+ (action
+  (with-outputs-to %%{target}
+   (run %s%%{dep:%s.exe})
+  )
+ )
+)
+
+(rule
+ (target %s.processed)
+ (action
+  (with-outputs-to %%{target}
+   (run ../strip_randomness.exe %%{dep:%s.actual})
+  )
+ )
+)
+
+|}
+    base expect_failure base base base
+
+let example_alias_stanza filename =
+  let base = Filename.remove_extension filename in
+  Fmt.pr
+    {|
+(alias
+ (name runtest)
+ (action
+   (diff %s.expected %s.processed)
+ )
+)
+|}
+    base base
+
+let is_example filename = Filename.check_suffix filename ".ml"
+
+let main expect_failure =
+  Sys.readdir "." |> Array.to_list |> List.sort String.compare
+  |> List.filter is_example
+  |> function
+  | [] -> () (* no tests to execute *)
+  | tests ->
+      global_stanza tests;
+      List.iter
+        (fun test ->
+          example_rule_stanza ~expect_failure test;
+          example_alias_stanza test)
+        tests
+
+open Cmdliner
+
+let expect_failure =
+  let doc =
+    Arg.info ~doc:"Negate the return status of the tests" [ "expect-failure" ]
+  in
+  Arg.(value & flag doc)
+
+let term =
+  Term.
+    (const main $ expect_failure, info ~version:"%%VERSION%%" "gen_dune_rules")
+
+let () = Term.(exit @@ eval term)
--- a/test/e2e/passing/basic.expected
+++ b/test/e2e/passing/basic.expected
@@ -1,0 +1,8 @@
+Testing suite-name.
+This run has ID `<uuid>`.
+ ...                test-a          0   First test case.[OK]                test-a          0   First test case.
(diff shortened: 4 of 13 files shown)


test/e2e/strip_randomness.ml
```diff
@@ -0,0 +1,37 @@
+let standardise_filesep =
+  let re = Re.compile (Re.str Filename.dir_sep) in
+  Re.replace_string ~all:true re ~by:"/"
```
NathanReb: Aren't you afraid this approach is a bit aggressive? I remember receiving notification about sanitizing test names so they might not contain `/` anymore in which case it's not much of an issue.

I don't think that's a major issue as there's probably no real harm in swapping those.

One thing worth considering is to proceed with this replacement based on the value of `Sys.os_type` or `Sys.win32` for instance.

Let me know what you think!

test/e2e/passing/basic.expected
```diff
@@ -0,0 +1,8 @@
...
+ ...                test-a          0   First test case.
[OK]                test-a          0   First test case.
+ ...                test-a          1   Second test case.
[OK]                test-a          1   Second test case.
+ ...                test-b          0   Third test case.
[OK]                test-b          0   Third test case.
+ ...                test-c          0   Fourth test case.
[OK]                test-c          0   Fourth test case.
+The full test results are available in `<build-context>/_build/_tests/00000000-0000-0000-0000-000000000000`.
+Test Successful in 0.000s. 4 tests run.
```
NathanReb: Might be worth replacing the execution time by a placeholder as well as it is something likely to change for more complex tests!

test/e2e/expect_failure.ml
```diff
@@ -0,0 +1,9 @@
+(* Run an executable and negate its return status *)
+let () =
+  let args = List.tl (Array.to_list Sys.argv) in
+  let command = Fmt.strf "%a" Fmt.(list ~sep:(const string " ") string) args in
+  let rel_command = Filename.(concat current_dir_name) command in
+  match Unix.system rel_command with
+  | WEXITED 0 -> exit 1
+  | WEXITED _ -> exit 0
+  | _ -> assert false
```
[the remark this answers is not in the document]
CraigFe: The remaining cases here are `WSIGNALLED` and `WSTOPPED`, both for external signals being passed into the binary-under-test, which I believe shouldn't be happening here under any circumstances?

Could still have a more explanatory error message here anyway, for internal documentation reasons.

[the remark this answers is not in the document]
NathanReb: Hmm right my bad, misread the cases!

Yeah I guess it would be more readable if the other cases were explicit!

test/e2e/failing/dune
```diff
@@ -0,0 +1,17 @@
+(include dune.autogen)
```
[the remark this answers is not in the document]
CraigFe: Yep, agreed. Done.

test/e2e/gen_dune_rules.ml
```diff
@@ -0,0 +1,88 @@
...
+ (target %s.actual)
+ (action
+  (with-outputs-to %%{target}
+   (run %s%%{dep:%s.exe})
+  )
+ )
+)
+
+(rule
+ (deps (:sanitiser ../strip_randomness.exe))
```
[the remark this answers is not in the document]
CraigFe: Nice, inferred dependencies are cool. Done.

test/e2e/strip_randomness.ml
```diff
@@ -0,0 +1,37 @@
...
+let uuid_replace =
+  let open Re in
+  let hex n = repn (alt [ rg 'A' 'F'; digit ]) n (Some n) in
+  let segmented_hex ns =
+    let segments = List.map (fun n -> [ char '-'; hex n ]) ns in
+    List.flatten segments |> List.tl |> seq
+  in
+  let t = segmented_hex [ 8; 4; 4; 4; 12 ] in
+  let re = compile t in
+  replace_string ~all:true re ~by:"00000000-0000-0000-0000-000000000000"
```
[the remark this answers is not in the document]
CraigFe: Agreed, though it should probably be `<uuid>` since this regexp doesn't actually ensure that the UUID is correct.

test/e2e/passing/basic.expected
CraigFe: Nice catch. Done.

test/e2e/strip_randomness.ml
CraigFe: I think this is currently safe, since these characters are banned in test names. In future, they may not be, but that would have to come with a more substantial revision of the structure of log outputs from Alcotest. 

I'm not particularly worried by this solution for now, but could be convinced otherwise -- bizarre characters in test names shouldn't be making their way onto the file-system whatever we do.

test/e2e/expect_failure.ml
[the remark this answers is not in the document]
CraigFe: Agreed; done.

test/e2e/strip_randomness.ml
NathanReb: It seems fine to me for now! Let's go with that!

samoht:
> I don't think the amount of boilerplate we need for this will feel like too much once we've added more tests and as we've discussed offline, I even think parts of this could be extracted into a cram-test-like framework or upstreamed to an existing one!

out of curiosity, did you consider using `mdx` for this?

samoht:
Looking good! Feel free to merge it whenever you think that's ready. That's a huge improvement over the current (empty) state.

NathanReb:
> out of curiosity, did you consider using mdx for this?

I hadn't but it's true we could make it work. I'm just not sure if that's a very good idea as it's a bit outside of `mdx`'s scope and I think it's better for `mdx` to focus on ensuring manuals or articles are correct than trying to also turn it into a testing framework.

I agree it wouldn't take much but I still think some features wouldn't quite fit. For instance, the random/non-deterministic output bits replacement logic we have here is quite different from what we want for `mdx` right now. Here we just want to replace it with a placeholder because we don't care about what it actually is as long as it matches what we expect it to be. In `mdx` we want a significant value in the output instead because that's clearer for the reader. I'm not quite sure how we can achieve both with a reasonable API and workflow. We could probably hack something together and make it work somehow but I'm afraid that by trying to be two different things at once the code base would suffer and we might regret it in the long run.

What I had in mind was something simpler than mdx, focused on end to end testing for CLI binaries similar to what dune uses for its blackbox tests. I was actually hoping we could extract that into a tool which we could use in both `dune` and `alcotest`.

We should probably also look into existing tools as there's probably one around!

test/e2e/failing/dune
```diff
@@ -0,0 +1,17 @@
+(include dune.inc)
+
+(rule
+ (targets dune.gen)
+ (deps (source_tree .))
+ (action
+  (with-stdout-to
+    %{targets}
+   (run %{dep:../gen_dune_rules.exe} --expect-failure)
```
NathanReb: Same as for `strip_randomness.exe`, the `{%dep:` can be removed as dune interprets the first `run` argument as a dependency in this case!

test/e2e/gen_dune_rules.ml
```diff
@@ -0,0 +1,87 @@
...
+   %a
+ )
+)
+|}
+    (pp_sexp_list Fmt.string) bases (pp_sexp_list Fmt.string) bases
+
+let example_rule_stanza ~expect_failure filename =
+  let base = Filename.remove_extension filename in
+  let expect_failure =
+    if expect_failure then "%{dep:../expect_failure.exe} " else ""
```
NathanReb: Same here, no need for the `{%dep:`!

samoht:
This is clearly off-topic, but `mdx` testing logic was extracted from dune blackbox tests :-) And merlin is using the alternate "craml" syntax of `mdx` to for its own [blackbox testing](https://github.com/ocaml/merlin/tree/master/tests) so this is definitely what `mdx` is designed for.

PR #181, opened 2019-08-26
Split module structure

CraigFe:
Splits the main Alcotest module into multiple files:

image

The only externally-visible change is to the location of the `Make` functor, which is now in `Core` instead. Since this functionality hasn't been released yet, this seems like fair game.

I think a very natural next step would be to pull out the `Cmdliner` interface to its own module, but wanted to avoid making significant changes to the logic for now.


samoht:
I am not very fond of having to use `Core.Make` to create a new backend but this is sufficiently rare to be fine.

CraigFe:
Not difficult to lift that `Make` functor to the top-level as well. My only reason for putting it into `Core` was that there's already quite a lot of stuff in the top-level.

samoht:
maybe just add an alias at top-level then?

Issue #120, opened 2018-01-06
Make alcotest compile to js


kennetpostigo:
Hey, I wanted to inquire if there is any interest in making alcotest compile to js? If there is then some points to make this possible:
- Include bs-native bsconfig.json in the repo (similar to jbuild)
-when compiling for js you must remove cmdliner as a dep and fmt (its cli module depends on cmdliner) and include a cli written in node.js in order to print to the console. The reason for removing cmdliner is because it depends on the stdlib Arg module, this specifically I don't believe can be compiled to js.

Khady:
> The reason for removing cmdliner is because it depends on the stdlib Arg module, this specifically I don't believe can be compiled to js.

I'm using `Arg` in a project compiled to javascript by bucklescript. So I think this part is not a problem.

kennetpostigo:
Interesting, not sure if there are any other Native only deps in Alcotest but if theres not, has anyone tried to use Alcotest to test bucklescript output?

ELLIOTTCABLE:
I think a lot of these BuckleScript-support issues across the ecosystem will improve heavily when Dune starts supporting BuckleScript (which is on the horizon, based on what I heard at ICFP.)

That said, that won't integrate well with `bsb` … which is its whole own problem, that can only really be fixed on the BuckleScript side. `/=`

dannywillems:
Is there any update on this? Could be nice to have a seamlessly integration with jsoo. We could imagine running the tests with NodeJS.

CraigFe:
@dannywillems: No explicit update on this, since noone has taken it on. Contributions are of course very welcome.

Regarding `js_of_ocaml` specifically, that should now be fairly easy to achieve thanks to @linse and @hannesm's work when creating the `alcotest-mirage` package. "Alcotest-on-JS" would just need to provide an implementation of [`Platform.MAKER`](https://github.com/mirage/alcotest/blob/master/src/alcotest-engine/platform.ml#L38) and then apply the [appropriate functors](https://github.com/mirage/alcotest/blob/master/src/alcotest-engine/core.mli#L88). I'm not aware as to whether the `cmdliner` dependency is an issue here, but since that is now implemented as a layer on top of `Core.S` it shouldn't be hard to deal with if so. Most of the difficulty would be in setting up appropriate CI to check that the `js_of_ocaml` builds work as expected.

[...]

PR #182, opened 2019-09-02
Make the CLI dependencies optional

CraigFe:
This refactors the CLI-related dependencies (`cmdliner`, `fmt.cli`) to an optional library: `alcotest.cli`. The motivation behind this change is to make it easier to compile to JS (https://github.com/mirage/alcotest/issues/120) and MirageOS.

 - All of the options that could previously be passed via CLI can now be passed to `run` and `run_with_args` (see [this collection of options](https://github.com/mirage/alcotest/pull/182/files#diff-7af91a1a38e4461394243052f899b25bR47)). This interface is simply a direct translation of the previous CLI one, so is perhaps not the easiest to use. The `run` and `run_with_args` functions exposed by the plain `Alcotest` library _must_ be configured in code, as they don't produce a CLI; __this is a breaking change__.

 - The `Alcotest_cli.{run, run_with_args}` functions keep the old CLI functionality of `Alcotest.{run, run_with_args}`.

 - As a side benefit, the `Alcotest_cli` run functions _also_ allow flags to be set in code (without the use of `argv`, which is somewhat clumsy). The CLI options override the code-configured ones where necessary.

While I was making this PR, it occurred to me that a consequence of sequencing `Lwt` and `Async` tests fully inside their respective concurrency monads (https://github.com/mirage/alcotest/pull/167) is that they can no longer be interacted with via the command line. This PR makes that issue explicit, since there are no `Alcotest_lwt.cli` or `Alcotest_async.cli` modules.  This seems like a shame, but perhaps there is a way around it?


src/cli/alcotest_cli.ml
```diff
@@ -0,0 +1,246 @@
...
+    compact : bool;
+    show_errors : bool;
+    quick_only : bool;
+    json : bool;
+    log_dir : string option;
+  }
+
+  let v_runtime_flags ~defaults (Verbose verbose) (Compact compact)
+      (Show_errors show_errors) (Quick_only quick_only) (Json json)
+      (Log_dir log_dir) =
```
Julow: Nice trick ! Would that work with polymorphic variants to avoid defining the types with dummy names ?

CraigFe: Good idea :slightly_smiling_face: Done.

CraigFe:
The `ocurrent` tests are failing for unrelated reasons: https://github.com/mirage/alcotest/issues/183.

samoht:
I am a bit reluctant to break the API for the main entry-point of the library. Maybe before going further, what are you trying to achieve here? Would it be possible/better to keep aliases in the `Alcotest` module but split the code in more specific modules?

It's probably better for a maintainability point of view if the modules are split, but from a user-point of view there should have been some benefits for breaking their API otherwise it'll be a hard sell.

CraigFe:
The main motivation for the PR is to decouple the core logic from the CLI interface, providing an intermediate `run` function whose configuration is entirely described by its type. In particular, this makes the logic easier to test. This could be achieved without breaking the API at all (`Alcotest` simply needs to include a tester built from `Cli` rather than `Core`).

Having cleanly abstracted the `Cmdliner` interface, it seems fairly natural to me to make it an optional dependency. This follows the conventions of `fmt` and `logs`, which also have optional `Cmdliner` wrappers. However, the biggest user benefit of this PR (being able to pass flags directly to the `run` function without going via `argv`) doesn't require an API break at all.

As I see it, the only benefit of the API break is dropping the Cmdliner dependency; perhaps this is not a good enough reason. I have no idea what proportion of Alcotest users use the CLI params / selective execution features.

samoht:
the only way to use alcotest today is to use the CLI interface so I am very wary of breaking things there. I think it would be better to keep the same API and provide extra hooks to allow to run the tests with a custom args and/or with different evaluation backend. 

Let's make the simple use-case easy to do and provide escape hatches for more exotic use-cases (mirage, js_of_ocaml, etc)

talex5:
Note that Mirage unikernels already include cmdliner anyway to parse their options (see the generated `key_gen.ml` files). And I think minimising dependencies is a bit less important for test dependencies like Alcotest anyway.

CraigFe:
There was some discussion of Cmdliner being an issue for compiling Alcotest to JS [here](https://github.com/mirage/alcotest/issues/120), but I haven't tested it. If this is the case, then providing the non-CLI interface for JS-only backends won't be very effective (since they'll be linking against Cmdliner anyway).

Having said that, I'm very happy with this PR just as a structural improvement to the code. We can deal with any dependency issues as and when they arise, and that will be much simpler once this PR has been merged. I'll fix the PR to re-export the CLI-compatible tester this evening.

CraigFe:
Done :slightly_smiling_face: I believe the only visible change to the `Alcotest.*` signature now is the extra optional arguments to the `run` and `run-with-args` functions.

aantron:
There is [`bs-cmdliner`](https://github.com/ELLIOTTCABLE/bs-cmdliner) on BuckleScript. I haven't tried it myself.

samoht:
Looks great, thanks for your patience :-)

Issue #183, opened 2019-09-06
Remove dependency on `tput`


CraigFe:
The `ocurrent` tests are failing becayse of a dependency on `tput` to determine the width of the TTY. This could be avoided by switching to a TIOCGWINSZ ioctl [as it is being done in Notty](https://github.com/pqwy/notty/blob/master/src-unix/native/winsize.c), or the terminal interactions could all be handled by adding a dependency to Notty directly.

emillon:
There's also https://github.com/cryptosense/terminal_size which exposes just this ioctl if you don't want to add a dependency to notty.

PR #185, opened 2019-09-16
Remove dependency on tput

CraigFe:
This fixes https://github.com/mirage/alcotest/issues/183 via @emillon's suggestion of adding a dependency to [terminal_size](https://github.com/cryptosense/terminal_size).

### DIFF
--- a/alcotest.opam
+++ b/alcotest.opam
@@ -21,6 +21,7 @@ depends: [
   "cmdliner"
   "uuidm"
   "re"
+  "terminal_size"
 ]
 
 synopsis: "Alcotest is a lightweight and colourful test framework"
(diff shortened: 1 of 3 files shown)


samoht:
`terminal_size` seems to be using `alcotest` for testing, so it might be a better idea to copy/paste the C bindings in a private lib instead of creating a cyclic dependency? tbh, I have no idea what will happen in opam with cyclic test dependencies!

CraigFe:
It also seems that this dependency can't be met on Windows (https://ci.appveyor.com/project/samoht/alcotest/builds/27443100#L225), which is problematic.

samoht:
Opam uses [this](https://github.com/ocaml/opam/blob/9c4ea9749b38ff60b2d3a3581ecf29fd55e16e28/src/core/opamStd.ml#L721-L765)

Maybe we don't care about the terminal size so much.

samoht:
See #189

PR #197, opened 2019-11-29
Make transitive dependencies explicit in dune files

CraigFe:


CraigFe:
Currently broken pre. 4.07 for similar reasons to https://github.com/mirage/index/pull/142. I still don't have any bright ideas to get it to work :slightly_frowning_face:

Issue #207, opened 2020-01-17
Environment variable ALCOTEST_COLOR


mjambon:
When running `dune runtest`, which runs my alcotest program with the default settings (color=auto), I don't get any color output. I'm not sure if it's a bug in dune or in `Fmt_tty` or in `Unix.isatty`. For consistency and for working around such problems in the future, I suggest adding an environment variable `ALCOTEST_COLOR` like we have for other options. [`Fmt_cli.style_renderer`](https://github.com/dbuenzli/fmt/blob/master/src/fmt_cli.mli) supports an `env` option, so it looks straightforward.

Current workaround for dune: only build the alcotest executable with `dune build`, then call it directly without using `dune runtest`.

CraigFe:
I believe the lack of colour by default is due to `dune`'s output buffering (https://github.com/ocaml/dune/issues/145), and as such this isn't really a bug. When I run with `dune runtest --no-buffer -j 1`, I get colours by default.

Having said that, I find the lack of colours by default frustrating and think adding an `ALCOTEST_COLOR` environment variable is good for consistency if nothing else :slightly_smiling_face: 

My preferred workaround currently is to use a desugared form of the `test` stanza in order to pass options to Alcotest:

```
(executable
 (name test)
 (modules test)
 (libraries alcotest))

(alias
 (name runtest)
 (package project)
 (action (run ./test.exe --color=always)))
```

We're [using this pattern in Irmin](https://github.com/mirage/irmin/blob/master/test/irmin-fs/dune#L6-L14) currently.

PR #209, opened 2020-01-17
Add support for environment variable ALCOTEST_COLOR

mjambon:
This offers an alternate way to specify color output, consistent with other options. See #207 . It just works. I don't think this needs a test since it's really just one call to an external function.


mjambon:
thank you!

PR #215, opened 2020-02-12
Add assert message to err only in case of failure

icristescu:
This prevents printing extra assert messages after running the tests. (you can see this behaviour before the fix, by running  `example/simple.ml`).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+- Remove the extra asserts printed sometimes after the results of a test. (#215,
+  @icristescu)
+
 ### 1.0.1 (2020-02-12)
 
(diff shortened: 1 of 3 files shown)


CraigFe:
Thanks. Can you add a test in `test/e2e/alcotest/passing` before this commit to show the behaviour being fixed? (and also write a CHANGES entry in an `Unreleased` section)

samoht:
I can confirm that this patch is fixing my issue (where a few `ASSERT` were printed on the stdout/stderr while running the tests). I'm not sure I understand anymore why this is the case, though :-)

samoht:
Closing in favour of #228

PR #216, opened 2020-02-13
Upgrade to Dune.2.0 and use opam file generation

CraigFe:
Also re-formats the `dune` files.


alcotest-async.opam
```diff
@@ -1,24 +1,31 @@
...
   "alcotest" {= version}
   "async_unix" {>= "v0.9.0"}
   "core_kernel" {>= "v0.9.0"}
 ]
-
-synopsis: "Async-based helpers for Alcotest"
+build: [
+  ["dune" "subst"] {pinned}
+  [
+    "dune"
```
samoht: this is ... ugly!

Issue #218, opened 2020-02-18
build failure with ocaml 4.04


olafhering:
This is just to get the correct info into opam files:

building 261448ec6d43d8ccba0da911d5cf80bd469e61b6 with ocaml.git#4.04 fails like this:
````
[   49s] + dune_for_release=--for-release-of-packages=alcotest,alcotest-lwt
[   49s] + dune installed-libraries
[   49s] astring                             (version: 0.8.3)
[   49s] astring.top                         (version: 0.8.3)
[   49s] bigarray                            (version: OCaml 4.04.2)
[   49s] bytes                               (version: OCaml 4.04.2)
[   49s] cmdliner                            (version: 1.0.4)
[   49s] compiler-libs                       (version: OCaml 4.04.2)
[   49s] compiler-libs.bytecomp              (version: OCaml 4.04.2)
[   49s] compiler-libs.common                (version: OCaml 4.04.2)
[   49s] compiler-libs.optcomp               (version: OCaml 4.04.2)
[   49s] compiler-libs.toplevel              (version: OCaml 4.04.2)
[   49s] dynlink                             (version: OCaml 4.04.2)
[   49s] fmt                                 (version: 0.8.5)
[   49s] fmt.cli                             (version: 0.8.5)
[   49s] fmt.top                             (version: 0.8.5)
[   49s] fmt.tty                             (version: 0.8.5)
[   49s] graphics                            (version: OCaml 4.04.2)
[   49s] logs                                (version: 0.7.0)
[   49s] logs.cli                            (version: 0.7.0)
[   49s] logs.fmt                            (version: 0.7.0)
[   49s] logs.lwt                            (version: 0.7.0)
[   49s] logs.threaded                       (version: 0.7.0)
[   49s] logs.top                            (version: 0.7.0)
[   49s] lwt                                 (version: 4.4.0)
[   49s] lwt.unix                            (version: 4.4.0)
[   49s] lwt_ppx                             (version: 1.2.4)
[   49s] lwt_react                           (version: dev)
[   49s] mmap                                (version: 1.1.0)
[   49s] num                                 (version: OCaml 4.04.2)
[   49s] num.core                            (version: OCaml 4.04.2)
[   49s] ocaml-migrate-parsetree             (version: 1.5.0)
[   49s] ocaml-migrate-parsetree.driver-main (version: 1.5.0)
[   49s] ocplib-endian                       (version: 1.0)
[   49s] ocplib-endian.bigstring             (version: 1.0)
[...]

PR #219, opened 2020-02-18
Tighten lower bound on Fmt for new result combinator

CraigFe:
Fixes https://github.com/mirage/alcotest/issues/218.


CraigFe:
I've also PRed these changes to `opam-repository` [here](https://github.com/ocaml/opam-repository/pull/15859).

Issue #225, opened 2020-03-12
Progressive test output broken with `--verbose` and newlines


CraigFe:
When run with the `--verbose` flag, if the test standard output contains a newline then this breaks the progressive result reporting in Alcotest (which relies on being able to use `\r` to overwrite an 'in progress' status). This can cause some of the test's output to be overwritten by Alcotest's output:

```ocaml
let () =
  Alcotest.run ~verbose:true "suite-name"
    [
      ( "alpha",
        [
          Alcotest.test_case "0 newlines" `Quick (fun () ->
              Format.printf "Print inside alpha");
        ] );
      ( "beta",
        (* Progressive result reporting is broken for this test, since the
           carriage return happens on the wrong line. *)
        [
          Alcotest.test_case "1 newline" `Quick (fun () ->
              Format.printf "Print inside beta\n");
        ] );
      ( "gamma",
        [
          (* Reporting is also broken here. Even worse, some of the test std.out
             is clipped by the eventual result of the test. *)
          Alcotest.test_case "1 newline + long line" `Quick (fun () ->
              Format.printf
                "Print inside gamma\n\
                 Lorem ipsum dolor sit amet, consectetur adipiscing elit, \
                 nullam malesuada dictum tortor in venenatis.");
        ] );
    ]
```

This example prints the following on a CR-sensitive terminal:

```
Testing suite-name.
This run has ID `<uuid>`.
[OK]                alpha          0   0 newlines.Print inside alpha
 ...                beta           0   1 newline.Print inside beta
[OK]                beta           0   1 newline.
 ...                gamma          0   1 newline + long line.Print inside gamma
[OK]                gamma          0   1 newline + long line.am malesuada dictum tortor in venenatis.
Test Successful in <test-duration>s. 3 tests run.
```

which is a complete mess :slightly_smiling_face:

PR #226, opened 2020-03-12
Add a reproduction of a bug with the --verbose flag

CraigFe:
This reproduces the bug detailed in https://github.com/mirage/alcotest/issues/225.

@samoht, this suggests that the `--verbose` flag is not in fact ignored. Can you give a minimal working example of the bug you were experiencing?

### DIFF
--- a/test/e2e/alcotest/passing/verbose_newlines.ml
+++ b/test/e2e/alcotest/passing/verbose_newlines.ml
@@ -26,4 +26,13 @@ let () =
                  Lorem ipsum dolor sit amet, consectetur adipiscing elit, \
                  nullam malesuada dictum tortor in venenatis.");
         ] );
+      ( "delta",
+        [
+          Alcotest.test_case "1 newline + long check" `Quick (fun () ->
+              Format.printf "Print inside delta\n";
+              Alcotest.(check unit)
+                "Lorem ipsum dolor sit amet, consectetur adipiscing elit, \
+                 nullam malesuada dictum tortor in venenatis."
+                () ());
+        ] );
     ]
--- a/test/e2e/alcotest/passing/verbose_newlines.ml
+++ b/test/e2e/alcotest/passing/verbose_newlines.ml
@@ -1,0 +1,29 @@
+(** Reproduction for https://github.com/mirage/alcotest/issues/225, testing the
+    interaction between `--verbose` and newlines in the test stdout. *)
+
+let () =
+  Alcotest.run ~verbose:true "suite-name"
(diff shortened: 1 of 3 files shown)


test/e2e/alcotest/passing/verbose_newlines.ml
```diff
@@ -0,0 +1,29 @@
...
+        [
+          Alcotest.test_case "1 newline" `Quick (fun () ->
+              Format.printf "Print inside beta\n");
+        ] );
+      ( "gamma",
+        [
+          (* Reporting is also broken here. Even worse, some of the test std.out
+             is clipped by the eventual result of the test. *)
+          Alcotest.test_case "1 newline + long line" `Quick (fun () ->
+              Format.printf
```
samoht: Can you add a few `Alcotest.check` tests too?

test/e2e/alcotest/passing/verbose_newlines.ml
```diff
@@ -0,0 +1,29 @@
+(** Reproduction for https://github.com/mirage/alcotest/issues/225, testing the
+    interaction between `--verbose` and newlines in the test stdout. *)
+
+let () =
+  Alcotest.run ~verbose:true "suite-name"
```
samoht: Is `~verbose` exactly the same as passing `--verbose` on the CLI?

CraigFe: It's not tested, but it should be. I can alter the test to work directly with `--verbose` instead.

CraigFe: With https://github.com/mirage/alcotest/pull/229, the two are tested to be the same.

test/e2e/alcotest/passing/verbose_newlines.ml
CraigFe: Added in https://github.com/mirage/alcotest/pull/226/commits/6478c96cd89e5129abbf0744d543af9a5d427aba.

This behaviour will change when https://github.com/mirage/alcotest/pull/228 is merged, since it's currently observing the `ASSERT` bug right now.

samoht:
Is this still needed? The mangled output doesn't look very good without `\r` ; do we want to fix this?

CraigFe:
Indeed, I'd intended https://github.com/mirage/alcotest/issues/225 (and the test in this PR) to track the mangled output coming from `--verbose` + `\r`. I think the `--verbose` + `ASSERT` issue is a separate matter.

PR #228, opened 2020-03-12
Fix interleaving of ASSERTs with other console output

CraigFe:
... by adding flushes in various places, being sure to flush internal `Format` buffers as well as those of the output channels.

This has the side-effect of ensuring that all `ASSERT` lines are captured in log files (when `--verbose` is not passed), which was previously not the case because of `Format` buffers being flushed after the redirection had stopped.

This supersedes the fix in https://github.com/mirage/alcotest/pull/215, which I now believe is taking the wrong approach: i.e. we should always print `ASSERTS` regardless of test failure, but should also be sure to capture them in the test logs correctly.

### DIFF
--- a/src/alcotest/core.ml
+++ b/src/alcotest/core.ml
@@ -324,8 +324,8 @@ module Make (M : Monad.S) = struct
     let open Suite in
     let test = suite.fn in
     let pp_event = pp_event t in
-    M.return () >>= fun () ->
     pp_event Fmt.stdout (`Start suite.path);
+    Fmt.(flush stdout) () (* Show event before any test stderr *);
     test args >|= fun result ->
     (* Store errors *)
     let () =
@@ -342,14 +342,18 @@ module Make (M : Monad.S) = struct
       in
       t.errors <- error @ t.errors
     in
+    (* Show any remaining test output before the event *)
+    Fmt.(flush stdout ());
+    Fmt.(flush stderr ());
     pp_event Fmt.stdout (`Result (suite.path, result));
     result
 
   let perform_tests t tests args = M.List.map_s (perform_test t args) tests
 
   let with_redirect file fn =
-    flush stdout;
-    flush stderr;
+    M.return () >>= fun () ->
+    Fmt.(flush stdout) ();
+    Fmt.(flush stderr) ();
     let fd_stdout = Unix.descr_of_out_channel stdout in
     let fd_stderr = Unix.descr_of_out_channel stderr in
     let fd_old_stdout = Unix.dup fd_stdout in
@@ -359,8 +363,8 @@ module Make (M : Monad.S) = struct
     Unix.dup2 fd_file fd_stderr;
     Unix.close fd_file;
     (try fn () >|= fun o -> `Ok o with e -> M.return @@ `Error e) >|= fun r ->
-    flush stdout;
-    flush stderr;
+    Fmt.(flush stdout ());
+    Fmt.(flush stderr ());
     Unix.dup2 fd_old_stdout fd_stdout;
     Unix.dup2 fd_old_stderr fd_stderr;
     Unix.close fd_old_stdout;
--- a/src/alcotest/core.ml
+++ b/src/alcotest/core.ml
@@ -324,6 +324,7 @@ module Make (M : Monad.S) = struct
     let open Suite in
     let test = suite.fn in
     let pp_event = pp_event t in
+    M.return () >>= fun () ->
     pp_event Fmt.stdout (`Start suite.path);
     Fmt.(flush stdout) () (* Show event before any test stderr *);
     test args >|= fun result ->
@@ -524,7 +525,9 @@ module Make (M : Monad.S) = struct
         Fmt.(pf stderr) "%a\n" Fmt.(list string) (List.rev error_acc);
         exit 1
     | Ok t -> (
-        ( Fmt.(pf stdout) "Testing %a.\n" bold_s name;
+        ( (* Only print inside the concurrency monad *)
+          M.return () >>= fun () ->
+          Fmt.(pf stdout) "Testing %a.\n" bold_s name;
           Fmt.(pf stdout) "This run has ID `%s`.\n" run_id;
           run_tests ?filter t () args )
         >|= fun test_failures ->
--- a/src/alcotest/core.ml
+++ b/src/alcotest/core.ml
@@ -324,6 +324,7 @@ module Make (M : Monad.S) = struct
     let open Suite in
     let test = suite.fn in
     let pp_event = pp_event t in
+    M.return () >>= fun () ->
     pp_event Fmt.stdout (`Start suite.path);
     test args >|= fun result ->
     (* Store errors *)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 ### Unreleased
 
 - Fix handling of CLI options for `Alcotest_{async,lwt}.run`. (#222, @CraigFe)
+- Fix interleaving of ASSERT outputs with the other test code, and ensure that
+  it is correctly captured in log files. (#215 #228, @icristescu @CraigFe)
 
 ### 1.0.1 (2020-02-12)
 
--- a/src/alcotest/alcotest.ml
+++ b/src/alcotest/alcotest.ml
@@ -140,7 +140,9 @@ let reject (type a) =
   (module M : TESTABLE with type t = M.t)
 
 let show_assert msg =
-  Format.eprintf "%a %s\n" Fmt.(styled `Yellow string) "ASSERT" msg
+  (* Flush any test stdout preceding the assert *)
+  Fmt.(flush stdout) ();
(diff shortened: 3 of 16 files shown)


samoht:
For the record, here is the difference between 0.8.3 and this branch on my use-case.

0.8.3:
```
$ dune exec -- test/functoria/test.exe test action 10 -v
Testing functoria.
This run has ID `B9EFFD6C-A823-4C6F-B706-E4316380F6E1`.
[SKIP]              cli              0   read_full_eval.
[SKIP]              cli              1   configure.
[SKIP]              cli              2   describe.
[SKIP]              cli              3   build.
[SKIP]              cli              4   clean.
[SKIP]              cli              5   help.
[SKIP]              cli              6   default.
[SKIP]              cli              7   generated_header.
[SKIP]              package          0   merge.
[SKIP]              package          1   pp.
[SKIP]              graph            0   var_name.
[SKIP]              graph            1   impl_name.
[SKIP]              graph            2   test_graph.
[SKIP]              action           0   sequence.
[SKIP]              action           1   sequence after error.
[SKIP]              action           2   bind passes the correct value to caller code.
[SKIP]              action           3   delete (file exists).
[SKIP]              action           4   delete (file does not exist).
[SKIP]              action           5   mkdir (new dir).
[SKIP]              action           6   mdkir (existing dir).
[SKIP]              action           7   mdkir (existing file).
[SKIP]              action           8   rmdir (non-existing dir).
[SKIP]              action           9   rmdir (existing dir).
 ...                action          10   rmdir (dir with contents).XXX YO
-----------------------------------------------------------------------------------------------------
ASSERT rmdir
-----------------------------------------------------------------------------------------------------
XXX YO
-----------------------------------------------------------------------------------------------------
ASSERT rmdir
-----------------------------------------------------------------------------------------------------
[OK]                action          10   rmdir (dir with contents).
[SKIP]              action          11   with_dir (create file).
[SKIP]              action          12   pwd (root).
[SKIP]              action          13   pwd (vfs).
[SKIP]              action          14   file exists (true).
[SKIP]              action          15   file exists (false).
[SKIP]              action          16   dir exists (true).
[SKIP]              action          17   file exists (true).
[SKIP]              action          18   file exists (false).
[SKIP]              action          19   dir exists (true).
[SKIP]              action          20   size_of (empty).
[SKIP]              action          21   size_of (error).
[SKIP]              action          22   size_of (large).
[SKIP]              action          23   run_cmd fails if the command doesn't exist.
[SKIP]              action          24   run_cmd succeeds if the command exists.
[SKIP]              action          25   write to nonexisting file.
[SKIP]              action          26   write to existing file.
[SKIP]              action          27   write.
Test Successful in 0.001s. 1 test run.
```

this branch:
```
$ dune exec -- test/functoria/test.exe test action 10 -v
Testing functoria.
This run has ID `5411185B-9E23-4EE5-98DE-72A69F78F82C`.
[SKIP]              cli              0   read_full_eval.
[SKIP]              cli              1   configure.
[SKIP]              cli              2   describe.
[SKIP]              cli              3   build.
[SKIP]              cli              4   clean.
[SKIP]              cli              5   help.
[SKIP]              cli              6   default.
[SKIP]              cli              7   generated_header.
[SKIP]              package          0   merge.
[SKIP]              package          1   pp.
[SKIP]              graph            0   var_name.
[SKIP]              graph            1   impl_name.
[SKIP]              graph            2   test_graph.
[SKIP]              action           0   sequence.
[SKIP]              action           1   sequence after error.
[SKIP]              action           2   bind passes the correct value to caller code.
[SKIP]              action           3   delete (file exists).
[SKIP]              action           4   delete (file does not exist).
[SKIP]              action           5   mkdir (new dir).
[SKIP]              action           6   mdkir (existing dir).
[SKIP]              action           7   mdkir (existing file).
[SKIP]              action           8   rmdir (non-existing dir).
[SKIP]              action           9   rmdir (existing dir).
 ...                action          10   rmdir (dir with contents).XXX YO
XXX YO
[OK]                action          10   rmdir (dir with contents).
[SKIP]              action          11   with_dir (create file).
[SKIP]              action          12   pwd (root).
[SKIP]              action          13   pwd (vfs).
[SKIP]              action          14   file exists (true).
[SKIP]              action          15   file exists (false).
[SKIP]              action          16   dir exists (true).
[SKIP]              action          17   file exists (true).
[SKIP]              action          18   file exists (false).
[SKIP]              action          19   dir exists (true).
[SKIP]              action          20   size_of (empty).
[SKIP]              action          21   size_of (error).
[SKIP]              action          22   size_of (large).
[SKIP]              action          23   run_cmd fails if the command doesn't exist.
[SKIP]              action          24   run_cmd succeeds if the command exists.
[SKIP]              action          25   write to nonexisting file.
[SKIP]              action          26   write to existing file.
[SKIP]              action          27   write.
Test Successful in 0.000s. 1 test run.
ASSERT rmdir
ASSERT rmdir
```

To reproduce I think you should try to call `Fmt.epr` in the code being tested and `Alcotest.check`.

CraigFe:
@samoht, @icristescu: this PR has now changed to supersede https://github.com/mirage/alcotest/pull/215 rather than just adding a test. It's ready for review.

src/alcotest/core.ml
```diff
@@ -345,7 +345,9 @@ module Make (M : Monad.S) = struct
     let open Suite in
     let test = suite.fn in
     let pp_event = pp_event t in
+    M.return () >>= fun () ->
     pp_event Fmt.stdout (`Start suite.path);
+    Fmt.(flush stdout) () (* Show event before any test stderr *);
```
samoht: why not printing stderr here too?

src/alcotest/core.ml
```diff
@@ -345,7 +345,9 @@ module Make (M : Monad.S) = struct
     let open Suite in
     let test = suite.fn in
     let pp_event = pp_event t in
+    M.return () >>= fun () ->
```
samoht: Why is this needed?

samoht:
It's a bit unclear to me where are the points we need flushing : but the current choices seem safe enough.

src/alcotest/core.ml
CraigFe: This ensures that the printing of the event occurs inside the concurrency monad (which wouldn't be the case otherwise), which is necessary for the Alcotest output to be correctly interleaved with the test output when using `Alcotest_lwt`. It's not tested to be necessary, but it could be.

src/alcotest/core.ml
CraigFe: There shouldn't be any pending `stderr` here: the `stderr` of each test case is flushed as soon as it finishes and all we've done since then is print events to `stdout`.

samoht:
Travis is reporting useful errors:

```
# File "src/alcotest/alcotest.ml", line 143, characters 13-19:
# Error: This expression has type Format.formatter
#        but an expression was expected of type out_channel
#     ocamlopt src/alcotest/.alcotest.objs/native/alcotest__Core.{cmx,o} (exit 2)
# (cd _build/default && /home/travis/.opam/ocaml-base-compiler.4.04.2/bin/ocamlopt.opt -w -40 -g -I 
```
You probably need a recent version of Fmt (`>=0.8.7`) to have `Fmt.flush`

CraigFe:
> You probably need a recent version of Fmt (`>=0.8.7`) to have `Fmt.flush`

Arf, thanks. Fmt bites me a lot with this sort of thing. Implemented it by hand in https://github.com/mirage/alcotest/pull/228/commits/3544af2d011955c5dd0b2b790752eaa565544a60 to avoid dropping support for OCaml 4.04.

CraigFe:
Not sure what to make of the latest CI error...

```
[ERROR] No switch ocaml-base-compiler.4.08.1 is currently installed. Did you
        mean 'opam switch create ocaml-base-compiler.4.08.1'?
        Installed switches are:

+opam switch create ocaml-base-compiler.4.08.1
```

(https://travis-ci.org/github/mirage/alcotest/jobs/661661199#L378)

src/alcotest/core.ml
icristescu: Do you mind adding the test which shows that it is necessary? (I don't understand how the alcotest outputs can occur inside the Lwt monad.) There aren't any tests using alcotest_lwt and that expose this bug anyway, it is maybe not a bad idea to add one anyway.

samoht:
I've restarted the failing job.

src/alcotest/core.ml
CraigFe: Sure, can do :slightly_smiling_face:

I added a test of the previous (broken) state in https://github.com/mirage/alcotest/pull/228/commits/30b47a7aa5a9bfb6a247288a0658895a777d4e21 (with an explanation). Note that the prints are observable but the `assert false` is not: the exception has been caught and hidden behind a bind point, so the failure isn't shown until the computation is run.

I fixed the behaviour in https://github.com/mirage/alcotest/pull/228/commits/847f2271a4cb99bb893189ff030b587b51c8076b. I went a step further and pushed the initial `"Testing ..."` message inside the concurrency monad too, for consistency.

CraigFe:
@samoht: good to go?

samoht:
Thanks!

PR #229, opened 2020-03-12
Refactor and clarify processing of CLI flags

CraigFe:
This cleans up the option handling logic and documents the precedence order that is expected to hold. This avoids setting default values for e.g.  `--verbose` in both `core.ml` and `cli.ml`.

Also adds a test verifying that the `--verbose` CLI flag takes priority over the compile-time parameter.


samoht:
LGTM

PR #232, opened 2020-03-12
Improve the documentation layout

CraigFe:
Recent features have made the docs a bit too complex for newcomers (particularly monad parameterisation). This undoes the damage by reordering to present simplest concepts first and adding some explanation here and there.

### DIFF
--- a/src/alcotest/cli.mli
+++ b/src/alcotest/cli.mli
@@ -25,6 +25,7 @@
 
 module type S = sig
   include Core.S
+  (** @inline *)
 
   val run :
     (?argv:string array -> string -> unit test list -> return) with_options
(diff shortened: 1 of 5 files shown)


src/alcotest/cli.mli
```diff
@@ -25,6 +25,7 @@
 
 module type S = sig
   include Core.S
+  (** @inline *)
```
samoht: What does `@inline` do? Is this an odoc attribute?

CraigFe: Yes, it's an Odoc attribute. It has the effect of concealing the `include Core.S` banner in the documentation (which seems good, since its really an implementation detail).

Without `@inline`:

image

With `@inline`:

image

Note that we can't do the same for the `Cli.S` include, since Odoc won't inline the type constraint `return = unit`.

I haven't found documentation for `@inline` anywhere; I just discovered it from usage in Jane Street packages to avoid adding documentation noise with a `_intf.ml` file (see https://github.com/janestreet/accessor/blob/master/src/subtyping.mli#L2). We could do this in a few places in the Irmin docs too...

samoht: That's really great! We need this for Irmin docs :-)

PR #233, opened 2020-03-13
Update CHANGELOG in preparation for 1.0.2 release

CraigFe:


CraigFe:
Actually, we should probably release at `1.1.0` since we're now depending on Dune 2.0.

PR #238, opened 2020-04-17
Make alcotest usable on Mirage

linse:
With this PR, we make alcotest usable in a MirageOS unikernel.
We moved the unix-specific code into the `alcotest.unix` sublibrary and defined the module signature `Platform` for code that depends on the platform. In there, we do 
- directory creation
- redirect stdout / sderr to files
- set the style renderer.

We implemented the opam package alcotest-mirage for MirageOS unikernel testing. An example is in  `test/e2e/alcotest-mirage`.

We would love your input on
- the semantics of `Alcotest.run`: no longer redirects output to files, independent of Unix - is that change a problem?
- `Alcotest_unix.run` behaves like `Alcotest.run` => should `Alcotest.run` be removed (API change)? 
- other approaches, such as deprecation warning on use of `Alcotest.run`?


CraigFe:
Thanks for the contribution! Mirage support is very fitting (and overdue).

The implementation looks good overall (though I have some quibbles such as a general lack of documentation for the new module types).

<hr>

I'm less keen about the naming of the package split (`alcotest`, `alcotest.lwt` & `alcotest-mirage`), for two reasons:
 - I think Dune dependencies and Opam packages should be kept in 1:1 correspondence unless strictly necessary, as I think the alternative is confusing for new users.
 - I suspect the change of semantics to `Alcotest.run` will be problematic for most users – personally I find the file redirection feature really important for any non-trivial codebase – and so I imagine the switch to `Alcotest_unix.run` generating a lot of downstream churn.

My counter-proposal is to split as follows:
 - `alcotest-kernel` (?): contains the platform-agnostic test runner (i.e. what `alcotest` is currently in this PR);
 - `alcotest-mirage`: instantiates the runner with Mirage + Lwt logic;
 - `alcotest`: instantiates the runner with Unix + Identity, providing exactly the same API / semantics as it currently does.

If I understand correctly, this split has the benefit of being invisible to all current users of Alcotest. It also avoids the question of what to do about the CLI params that don't make sense without file redirection (`--tail-errors`, `--show-errors`, `--verbose` and to some extent `--compact`).

samoht:
Thanks for your patch!

A very quick review: I like the approach of the patch (abstracting away the platform specific bits). However I am not fond of changing UX regarding `Alcotest.run` that 99% of our users will be using. Ideally, this function should not change (and be an alias for `Alcotest_unix.run`) without brining unix dependencies to `alcotest-mirage`. Maybe this could be done by re-arranging the modules a bit?

Lupus:
Dune virtual libraries might be a good fit to solve this problem. In this case `Alcotest.run` would remain and maintain existing behavior, unix could be listed as default implementation, which is also backwards-compatible, yet other implementations can be plugged in to support mirage, js, other platforms.

hannesm:
Hi,

we improved this PR as you requested. Now, there is a `alcotest-core` opam package which does not depend on Unix. The `alcotest` API is unmodified (apart from the functors which are only exported via `alcotest-core`).

The AppVeyor error is unrelated (`curl` is missing on cygwin); the OCaml-CI error in 4.06.0 is an OCaml outputting change (which is different in newer OCaml versions -- `Alcotest_core.Core` vs `Alcotest_core__Core`) - we're not sure how to fix this.

We hope this version works for you and can be merged. If not, ping us.

samoht:
Thanks for your changes. I'll review this more carefully next week, but I don't think that adding a new package (`alcotest-core`) is needed.

hannesm:
CI passes, we needed to include `Sys.getcwd` into platform. would be great to get this reviewed and merged @CraigFe @samoht

samoht:
LGTM

@CraigFe you mentioned this could break for projects using `(implicit_transitive_deps false)`. Do you have an example in mind?

CraigFe:
> @CraigFe you mentioned this could break for projects using `(implicit_transitive_deps false)`. Do you have an example in mind?

@samoht: for instance, the type of test filters is no longer observable without explicitly including `alcotest.core` in the libraries stanza:
```
ᐅ dune exec test/test.exe
File "test/test.ml", line 269, characters 36-48:
269 |     Alcotest_lwt.run ~filter:(None, IntSet.empty) "test" [
                                          ^^^^^^^^^^^^
Error: This expression has type IntSet.t
       but an expression was expected of type
         Alcotest_core__.Core.IntSet.t option
```

On closer inspection, this may be the only such case (in which case it's hardly an issue).

The greater breakage is probably that the `Alcotest.Cli` and `Alcotest.Core` modules no longer exist: this is a [breaking change for OCurrent](https://github.com/ocurrent/ocurrent/blob/master/test/test.ml#L264) (albeit only because it was working around a bug that has now been fixed).

hannesm:
> The greater breakage is probably that the Alcotest.Cli and Alcotest.Core modules no longer exist: this is a breaking change for OCurrent (albeit only because it was working around a bug that has now been fixed).

I'm not a fan of re-exporting modules, that's why they no longer exist - if you re-export them and then use the functors with the wrong paths, this leads to compilation errors. I also think of them as internal modules, and unlikely being used outside of alcotest.

> alcotest.core vs alcotest-core

I don't mind either.

CraigFe:
> I'm not a fan of re-exporting modules, that's why they no longer exist - if you re-export them and then use the functors with the wrong paths, this leads to compilation errors. I also think of them as internal modules, and unlikely being used outside of alcotest.

I don't think it's a case of _re-exporting_ modules so much as pre-applying them with the `Unix` platform. I wouldn't suggest providing those helpers if it weren't a matter of backwards compatibility with the functors that were there before. Perhaps @samoht has some thoughts on this.

<hr/>

I've PRed into this one with some proposed changes (see https://github.com/roburio/alcotest/pull/1). Copying the PR description from there:

- rename `alcotest.core` to `alcotest.engine` to avoid any confusion with JS packages (I think this is a better descriptor anyway);

- cleanup the Mirage `Platform` functor to be slightly less brittle (avoid reference to `Pp.run_result`), and use this to hide some internal modules inside a `Private` wrapper;

- provide test runner functors with `Unix` pre-applied (this keeps some degree of backwards compatibility);

- add a CHANGES entry.

In my opinion, with those changes the PR can be merged. Don't let that block further discussion if you disagree with my proposals.

CraigFe:
Will merge as https://github.com/mirage/alcotest/pull/244 when the CI passes there (can't push to this branch). It contains a test fix that I forgot to include in my branched PR.

hannesm:
@CraigFe thanks for your PR, we merged it. This is good to be merged now.

CraigFe:
Thanks for the contribution :blush:

PR #239, opened 2020-04-30
labeled version `check_l` of `check` exptected vs. actual

hartmut27:
additional function `check_l` as the labeled arguments version of `check`

### DIFF
--- a/src/alcotest/alcotest.mli
+++ b/src/alcotest/alcotest.mli
@@ -111,6 +111,9 @@ val reject : 'a testable
 val check : 'a testable -> string -> 'a -> 'a -> unit
 (** Check that two values are equal. *)
 
+val check' : 'a testable -> msg:string -> expected:'a -> actual:'a -> unit
+(** Check that two values are equal (labeled variant of {!check}). *)
+
 val fail : string -> 'a
 (** Simply fail. *)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ### Unreleased
 
+- Add `Alcotest.check'`, a variant of `Alcotest.check` with labeled arguments.
+  (#239, @hartmut27)
+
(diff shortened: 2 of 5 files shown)


src/alcotest/alcotest.mli
```diff
@@ -111,6 +111,9 @@ val reject : 'a testable
 val check : 'a testable -> string -> 'a -> 'a -> unit
 (** Check that two values are equal. *)
 
+val check_l : t:'a testable -> msg:string -> expected:'a -> actual:'a -> unit
```
CraigFe, line 114:
```suggestion
val check_l : 'a testable -> msg:string -> expected:'a -> actual:'a -> unit
```
I'm not sure this first label is adding much in the way of code-descriptiveness – `t` values in other label-heavy libraries like Base and Core are generally not labelled. 

Perhaps we should drop it? It will still be possible to pass all four arguments in any order.

CraigFe, line 114:
```suggestion
val check' : t:'a testable -> msg:string -> expected:'a -> actual:'a -> unit
```
I'd normally expect such "syntax alternatives with equivalent behaviour" to be primed: slightly less descriptive, slightly easier to type. What do you think? (Perhaps @samoht has an opinion.)

src/alcotest/alcotest.mli
```diff
@@ -111,6 +111,9 @@ val reject : 'a testable
...
+(** Check that two values are equal, labeled version *)
```
CraigFe, line 115:
```suggestion
(** Check that two values are equal, labeled version. *)
```

src/alcotest/alcotest.mli
hartmut27: i like the primed check function too :)

CraigFe:
Have rebased and added a CHANGES entry & test.

Merging now. Thanks :slightly_smiling_face:

PR #241, opened 2020-05-03
Fix test-case name validation regex.

### LINKED ISSUE #240: Test suite fails if there is a path separator in user-supplied strings.
It looks like Alcotest tries to use the test name verbatim to create the output file : 
```
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "open", "/home/mefyl/dev/routine/mellifera/_build/default/tests/_build/_tests/8520A9D5-E675-42A0-9B41-883792EA0F62//some/hardcoded/path.000.output")
```

Which fails since parent directories (`some/hardcoded`) do not exist. Note that incidentally a leading slash works.

I'm keen on submitting a patch if the fix is agreed on. Do we want to escape the test names ?

mefyl:
Fix #240.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### Unreleased
+
+- Fix bug with validation of test names containing path separators (these break
+  Alcotest when attempting to generate a corresponding log file). The accepted
+  characters are: alphanumerics, underscore, hypen and space. (#241, @mefyl)
+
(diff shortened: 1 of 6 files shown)


CraigFe:
Neat. It took me a while to see what the diff was doing :grin:

Thanks for the fix and the (belated) test. Have added a CHANGES entry.

PR #245, opened 2020-05-22
Mitigate race conditions in Unix platform implementation

CraigFe:
See https://github.com/mirage/alcotest/pull/242#issuecomment-631109239.

The test suite creates contention for these symlinks, so we were occasionally getting failures due to these race conditions. This just adds a few retries to each operation on failure.

### DIFF
--- a/src/alcotest/alcotest.ml
+++ b/src/alcotest/alcotest.ml
@@ -36,18 +36,43 @@ module Unix (M : Alcotest_engine.Monad.S) = struct
 
   let getcwd = Sys.getcwd
 
+  let unlink_if_exists file =
+    let rec inner ~retries =
+      try Unix.unlink file with
+      | Unix.Unix_error (Unix.ENOENT, _, _) -> ()
+      | Unix.Unix_error (Unix.EINTR, _, _) ->
+          if retries > 5 then
+            Fmt.failwith "Failed %d times to unlink file %s (Unix.EINTR)."
+              retries file
+          else inner ~retries:(retries + 1)
+    in
+    inner ~retries:0
+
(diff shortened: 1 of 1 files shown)


src/alcotest/alcotest.ml
```diff
@@ -36,18 +36,43 @@ module Unix (M : Alcotest_engine.Monad.S) = struct
...
+
+  let symlink ~to_dir ~target ~link_name =
+    let rec inner ~retries =
+      try Unix.symlink ~to_dir target link_name
+      with Unix.Unix_error (Unix.EEXIST, _, _) ->
+        if retries > 5 then
+          Fmt.failwith "Failed %d times to create symlink %s (Unix.EEXIST)"
+            retries target
+        else (
+          unlink_if_exists link_name;
```
icristescu, line 59: I'm wondering if you need to unlink it here? This is called after the `unlink_if_exists`, so presumably that succeeded. If the symlink is already present at this point, then it is linking to the right `dir`, right?

CraigFe: When running inside Dune (as is the case for Alcotest's own set of tests), it's possible that either symlink can be created _after_ the first call to `_unlink_if_exists` but _before_ we call `symlink`. So it's still possible that we'll get `EEXIST` at this point and have to do another deletion.

I considered not bothering with the initial delete at all and just letting it be handled as part of this retry loop, but I suspect it's still better to have the explicit unlink to make it clearer what's going on in the sequential case.

icristescu: I see, thanks.
And I agree on having the first deletion explicit.

PR #246, opened 2020-05-25
Formatting improvements

CraigFe:
Fixes https://github.com/mirage/alcotest/issues/252 and fixes https://github.com/mirage/alcotest/issues/240.

No API changes, but lots of diffs to pretty-printing logic.

image

Shortlist of changes:
- output is generally more colorful (e.g. passing around `unit Fmt.t` values rather than `string`s);
- improve diff shown on `Alcotest.check` failure to be better aligned & spaced;
- highlight the test case whose error is printed with `>`;
- keep a more consistent distinction between `ERROR` and `FAIL` states;
- show shorter directory names to the user (e.g. use the symlinks rather than UUIDs, collapse home directory to `~`).


samoht:
What's the difference between `FAIL` and `ERROR`?

CraigFe:
- `FAIL`: test failures caused by `Alcotest.fail` and `Alcotest.check_raises`;
- `ERROR`: test failures caused by `Alcotest.check` or exceptions thrown in test.

It's fairly arbitrary, but I'm preserving the existing distinction. Perhaps we want to reconsider it at this stage.

CraigFe:
Fixes https://github.com/mirage/alcotest/issues/252.

samoht:
Yes we should drop the distinction. I can't remember why I thought it was a good idea initially :p

CraigFe:
> Yes we should drop the distinction.

@samoht: Now done.

New changes:

- Always use `FAIL` rather than `ERROR`
- Following discussion in https://github.com/mirage/alcotest/pull/249, display any UTF-8 characters in test-case names in unescaped form.

Also trying to sneak in some general refactoring in the internal handling of data. Mostly pushing definitions and logic into a new `model.ml` file that keeps some better abstraction on e.g. whether strings have been escaped, which form should be pretty-printed etc.

Happy to pull these changes into a separate PR if requested.

CraigFe:
Rebased. This can now be considered for merge.

test/e2e/alcotest/failing/check_basic.expected
```diff
@@ -1,24 +1,160 @@
...
-[failure] Error bool: expecting
-true, got
-false.
-
-The full test results are available in `<build-context>/_build/_tests/<uuid>`.
-13 errors! in <test-duration>s. 13 tests run.
+Testing `check_basic'.
+This run has ID `<uuid>'.
+
+  ...           different basic              0   bool.ASSERT bool
```
samoht, line 4: why is there an `ASSERT bool`in the middle of that line?

test/e2e/alcotest/failing/check_basic.expected
```diff
@@ -1,24 +1,160 @@
...
+│ [FAIL]        different basic              0   bool.                         │
+└──────────────────────────────────────────────────────────────────────────────┘
+FAIL bool
+
+   Expected: true
+   Received: false
+ ──────────────────────────────────────────────────────────────────────────────
+
+
+┌──────────────────────────────────────────────────────────────────────────────┐
```
samoht, line 41: Is the new default behaviour to show all the failures? I don't think that's a good idea, the previous default is just to show the first one unless `ALCOTEST_SHOW_ERRORS` is set (if I remember correctly).

src/alcotest-engine/alcotest_engine.mli
```diff
@@ -51,3 +51,6 @@ module Private : sig
   module Utils = Utils
   module Pp = Pp
 end
+
+module Higher = Higher
```
samoht, line 55: Do we really need to use `Higher`? This brings a new level of complexity which might make future contributions and maintenance a bit more difficult that it is already.

test/e2e/alcotest/failing/check_basic.expected
CraigFe: No, this isn't default. There's a corresponding change to the `Alcotest.run` to pass `~verbose:true` (https://github.com/mirage/alcotest/pull/246/files#diff-fb43ad350d97f3cf0e25d21ec53a881dR7).

I felt that this is a better test if all of the errors are printed.

src/alcotest-engine/alcotest_engine.mli
CraigFe: We don't. I like that it avoids the need for `Platform.MAKER` in favour of `Platform.S`, but there's not a huge benefit.

CraigFe:
Rebased, and switched out the last change for one that doesn't use brands.

CraigFe:
Pushed a small addition to the UTF-8 improvement that also handles printing of UTF-8 characters in _suite_ names (previously we were just dealing with test names). This now fixes https://github.com/mirage/alcotest/issues/240.

Now is a good time for a second review :slightly_smiling_face:

samoht:
This looks good, thanks!

Issue #250, opened 2020-05-26
Documentation about alcotest tests


gs0510:
While writing alcotest tests, I saw diffs on the terminal, but copy-pasting the diffs to expected result did not work. I did not know about `dune promote` to update the diff, and it might be useful to add documentation. Maybe in `Contributing.md`?

CraigFe:
Your review on https://github.com/mirage/alcotest/pull/251 is welcomed :blush:

Issue #124, opened 2018-04-18
Review the slow/quick distinction


samoht:
It could be useful to have something more powerful, with custom tags and a simple query language. Also, on simple projects nobody care about that feature so we should not make it mandatory.

emillon:
A way to do that is to group the slow tests in a different executable, so that one can use `dune runtest tests/fast/` to only run fast ones (or do not add the slow ones to `@runtest`).

PR #251, opened 2020-05-26
Add a CONTRIBUTING.md file

CraigFe:
... explaining the new high-level project structure and how to work with the test suite.

Resolves https://github.com/mirage/alcotest/issues/250.


CraigFe:
This depends on https://github.com/mirage/alcotest/pull/248.

gs0510:
LGTM! :tada:

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+
+If you change the Alcotest output in some way, you will likely need to update
+the expect test snapshots:
+
+1. run `dune test` to get a list of diffs to the tests,
+1. check that the diffs are as expected,
+1. run `dune promote` to update the snapshots in the repository.
+
+To add a new test, create a new `foo.{ml,expected}` pair in one of the test
+directories and run `dune test --auto-promote` to automatically generate the
```
gs0510, line 83: How does one use the quick/slow test feature with dune? 

I tried it with both `dune test` and `dune runtest` but couldn't figure it out.

CraigFe: It's possible to do it with `ALCOTEST_QUICK_TESTS=true dune test`, but that's not particularly ergonomic :wink:. We probably shouldn't have this hard-coded into our API, since it seems like a fairly special-case feature to me (c.f. https://github.com/mirage/alcotest/issues/124).

gs0510: Ah cool! Does this mean there isn't an ergonomic solution for filtering tests via `dune (run)test`?

CraigFe: There is not.

I generally do `dune exec test/foo/test.exe -- test <filter>` when I want to do this, which also works to pass CLI options but only runs a single test binary at a time.

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+When running an expect test, these portions of the output are sanitized by
+[`strip_randomness.ml`](./test/e2e/strip_randomness.ml) before being compared to
+the `.expected` file, resulting in files with portable placeholders:
+
+```
+The full test results are available in `<build-context>/_build/_tests/<uuid>`.
+Test Successful in <test-duration>s. 4 tests run.
+```
+
+If your change adds non-determinism to the Alcotest output, you may need to
```
gs0510, line 104: It might be useful to add the bit about adding changelog.

CraigFe: Done :slightly_smiling_face:

gs0510:
LGTM!! :tada:

PR #253, opened 2020-05-30
Add Ocaml `bytes` testable.

mefyl:
Printing relies on `Format` escaping with `%S`, which I think is exactly what we want : plain text as long as it's printable, escaped otherwise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,4 +10,4 @@ clean:
 	dune clean
 
 format:
-	ocamlformat --inplace $(git ls-files '*.ml' '*.mli')
+	dune build --auto-promote @fmt
--- a/Makefile
+++ b/Makefile
@@ -9,4 +9,5 @@ test:
 clean:
 	dune clean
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -9,4 +9,5 @@ test:
 clean:
 	dune clean
 
-.PHONY: all clean
+format:
+	ocamlformat --inplace $(git ls-files '*.ml' '*.mli')
```
CraigFe, line 13: Is this the same as `dune build --auto-promote @fmt`? If so, I prefer that.

mefyl: Certainly, I didn't know about it

mefyl:
I think https://travis-ci.org/github/mirage/alcotest/jobs/692835567 is transient (GPG key error), is there any way to rerun that one ?

CraigFe:
I've restarted it.

CraigFe:
Thanks :)

PR #263, opened 2020-07-23
Add a CI check for changelog entries

CraigFe:
... using a GitHub Actions script implemented by @emillon (https://github.com/ocaml-ppx/ocamlformat/blob/6c2524ba7cb737dda647176f47f155b4d74aa90f/.github/workflows/main.yml).

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -1,0 +1,20 @@
+name: Changelog check
+
+on:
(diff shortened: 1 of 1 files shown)


emillon:
FYI, if you're wondering why it's not running in this PR: it will always run the workflow as present in the target repository. So, if you want to see it working from the PR itself, the PR needs to come frome the mirage repo itself.

Issue #264, opened 2020-08-03
uncaught exception: Unix.EPERM unlink


hannesm:
hello,

when running alcotest (1.2.1) on the most recent icalendar release (v0.1.3), I run into the following issue:
```bash
$ dune runte
        test alias test/runtest (exit 125)
(cd _build/default/test && ./test.exe)
test.exe: internal error, uncaught exception:
          Unix.Unix_error(Unix.EPERM, "unlink", "/usr/home/hannes/devel/mirage/icalendar/_build/default/test/_build/_tests/")
          Raised by primitive operation at file "src/alcotest/alcotest.ml", line 40, characters 10-26
          Called from file "src/alcotest/alcotest.ml", line 69, characters 8-33
          Called from file "src/alcotest-engine/core.ml", line 348, characters 4-104
          Called from file "src/alcotest-engine/core.ml", line 447, characters 4-365
          Called from file "src/alcotest-engine/cli.ml", line 117, characters 4-128
          Called from file "cmdliner_term.ml", line 25, characters 19-24
          Called from file "cmdliner.ml", line 117, characters 32-39
Testing `'.
This run has ID `23AD5240-A1C0-486A-86C0-5E64DE247749'.


```

The issue is resolved by:
```OCaml
diff --git a/test/test.ml b/test/test.ml
index 0bf0f7b..7372d3b 100644
--- a/test/test.ml
+++ b/test/test.ml
@@ -2813,4 +2813,4 @@ let tests = [
 
 let () =
   Printexc.record_backtrace true;
-  Alcotest.run "" tests
+  Alcotest.run "icalendar suite" tests
```

I thought it is worth reporting, eventually you'd like to ensure that the suite named `""` does not entail any issues (it was not an issue on a previous run, i.e. an earlier version of alcotest - take a look at the travis logs of roburio/icalendar if you want to figure out which alcotest version worked previously).

CraigFe:
Thanks for the report. You're right, we should have empty suite names and empty test names in our own test suite.

As far as resolving this issue goes, I'm tempted to say that empty suite names and test names should be unsupported (and raise a much nicer error message). They cause issues for both the console output and the log capturing mechanism, as shown here. What do you think?

hannesm:
@CraigFe I don't have a particular opinion on this, since my mental model of allowed test and suite names is rather brittle (I remember there were changes disallowing the same test / suite name twice, and something in respect to unicode, plus file path special characters (`/`, `..`)).

[...]

PR #265, opened 2020-08-03
Emit an error when the user supplies an empty suite name

CraigFe:
Fix https://github.com/mirage/alcotest/issues/264.

This emits an error message when the user supplies an empty test suite name, which previously raised an uncaught exception due to trying to create a corresponding symlink with an empty name. As discussed in the linked issue, it seems better to avoid empty suite names altogether.

Also adds a test for the similar case for empty _test_ names, which isn't as problematic since we already append integers to these files on disk. In this case, I just chose to alter the filenames slightly to make sure that they don't start with a `.`.

Introduces a common `user_error` function for handling such cases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
-### unreleased
+### Unreleased
+
+- Fail gracefully when the user supplies an empty suite name. (#265, @CraigFe)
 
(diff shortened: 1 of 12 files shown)


CraigFe:
Will merge at the end of this week unless there are objections.

PR #266, opened 2020-08-10
Explicitly declare Alcotest's deps on fmt

NathanReb:
It was transitively working thanks to fmt.tty with the upstream fmt version but trying to use fmt's dune port failed.


CraigFe:
Ta :slightly_smiling_face: `(implicit_transitive_deps false)` bites me again...

Issue #272, opened 2020-09-03
spurious test suite failures in automated builds


olafhering:
The [ocaml-alcotest](https://build.opensuse.org/package/show/devel:languages:ocaml/ocaml-alcotest) runs `dune runtest --verbose --for-release-of-packages=alcotest,alcotest-lwt` after each automated rebuild.

This `dune runtest` randomly fails. Triggering a rebuild will result in a successful run.

The default output provides usually no hint what went wrong.

Examples of failure:
````
(SLE_12/ppc64le)
[   85s] Command [216] exited with code 125:
[   85s] $ (cd _build/default/test/e2e/alcotest/passing && ./quick_only_regex.exe &> _build/default/test/e2e/alcotest/passing/quick_only_regex.actual
````
````
(SLE_15/ppc64le)
[   73s] Command [255] exited with code 1:
[   73s] $ (cd _build/default && /usr/bin/diff -u test/e2e/alcotest/failing/invalid_arg_in_test.expected test/e2e/alcotest/failing/invalid_arg_in_test.processed)
[   73s] --- test/e2e/alcotest/failing/invalid_arg_in_test.expected	2020-09-01 18:07:25.760000000 +0000
[   73s] +++ test/e2e/alcotest/failing/invalid_arg_in_test.processed	2020-09-01 18:07:34.980000000 +0000
[   73s] @@ -8,7 +8,8 @@
[   73s]  ┌──────────────────────────────────────────────────────────────────────────────┐
[   73s]  │ [FAIL]        test-a          1   Failing.                                   │
[   73s]  └──────────────────────────────────────────────────────────────────────────────┘
[   73s] -[invalid] Failing test
[   73s] +ASSERT Failed
[   73s] +FAIL Failed
[   73s]  
[   73s]  Logs saved to `<build-context>/_build/_tests/<test-dir>/test-a.001.output'.
[   73s]   ──────────────────────────────────────────────────────────────────────────────
````
````
(openSUSE_Tumbleweed/aarch64)
[   96s] Command [257] exited with code 1:
[   96s] $ (cd _build/default && /usr/bin/diff -u test/e2e/alcotest/failing/long_test_case_name.expected test/e2e/alcotest/failing/long_test_case_name.processed)
[   96s] --- test/e2e/alcotest/failing/long_test_case_name.expected	2020-09-01 17:22:48.590000000 +0000
[...]

PR #273, opened 2020-09-03
Assign all test suites distinct names

CraigFe:
This avoids a bug in which running the tests concurrently would occasionally result in spurious failures, due to different binaries attempting to write to the same log files.

Partial fix for https://github.com/mirage/alcotest/issues/272.


CraigFe:
As mentioned in #272, we could add some sort of locking mechanism to prevent users from making the same mistake as we did here. When capturing output in a directory, Alcotest would first try to write a lock file to that directory. This probably isn't such a big issue outside Alcotest itself, since users will tend to have far fewer test binaries, but it's worth considering. It's slightly tricky as we need to consider the edge-cases of when to delete the lock-file.

hhugo:
I've the same issue, I believe we should not use the "pretty" dir name (including the testsuite name)  when writing files to disk.

PR #278, opened 2020-09-11
async: extract_exn

vbmithr:
This is need in order for the Async version to work properly.
Without it, `try_with` with not return the exception raised by user, but instead an exception wrapper which does not play nice with testing.

### DIFF
--- a/src/alcotest-async/alcotest_async.ml
+++ b/src/alcotest-async/alcotest_async.ml
@@ -11,7 +11,9 @@ module Tester =
       let bind x f = bind x ~f
 
       let catch t on_error =
-        try_with t >>= function Ok a -> return a | Error exn -> on_error exn
+        try_with ~extract_exn:true t >>= function
+        | Ok a -> return a
+        | Error exn -> on_error exn
     end)
 
 include Tester


CraigFe:
Not being familiar with Async, my understanding is that `extract_exn:true` discards some backtrace and monitor information that might be helpful for debugging. In what sense does the exception wrapper not play nicely with testing?

vbmithr:
In the sense that `alcotest` (apparently) uses exceptions when failing tests and uses the exception to display the problem (i.e. the function returned x instead of y in the case of an int function being tested). If not using extract_exn:true, this information is lost and you just see an unhelpful general exception.

CraigFe:
Added a test case for this in https://github.com/mirage/alcotest/pull/279.

It's true that Async's exception wrapping prevents exceptions raised by `Alcotest.check` from being properly displayed at the top-level, and adding `extract_exn:true` fixes this behaviour. Unfortunately, this also throws away stack trace information from user exceptions in tests ([example](https://github.com/mirage/alcotest/pull/279/files#diff-5645d8163fb72fc8b3d4b48692e80191R32)).

I think a more granular solution is necessary to ensure that only [`Check_error`](https://github.com/mirage/alcotest/blob/1c51b8ffa213e929740578dc6acac8e3586c2eb9/src/alcotest-engine/core.mli#L17) exceptions are unwrapped in this way.

Issue #277, opened 2020-09-11
Unicode boxes do not track terminal width


emillon:
Hi,

On a terminal 85 columns wide, the default behavior outputs boxes that are not readable:

![alcotest](https://user-images.githubusercontent.com/496345/92930424-576cfa00-f442-11ea-834c-7428be7e0452.png)

That's related to #72 but I think that's not the whole story because it seems that alcotest now assumes that the terminal is 80 columns wide, which would fit.

Thanks!

CraigFe:
Indeed, the more fundamental problem here is that we're not yet using `TIOCGWINSZ`. With a stub for that, we could fix both this and #72 very easily.

emillon:
But I think there's something off in the calculation too, otherwise something that assumes a 80 column terminal would work fine on a 85 column terminal, wouldn't it?

CraigFe:
Indeed. The formatting doesn't try very hard to meet an 80 column width at the moment (because this is already an arbitrary choice). In particular, long test names are printed as-is with no wrapping (as in [this test case](https://github.com/mirage/alcotest/blob/master/test/e2e/alcotest/failing/long_test_case_name.expected#L7-L9), which is similar to the example you've given), with the idea being that the user currently knows better than Alcotest how wide they want their output.

There's nothing particularly challenging about providing a guarantee of "no lines longer than terminal width", but IMO not much point in doing that until we're actually measuring terminal width.

Issue #72, opened 2016-09-14
test name output longer than a terminal line looks strange


hannesm:
```
[OK]                RepositoryReleases               2   not authorised.
[OK]                RepositoryReleases               3   bad release.
 ...                RepositoryReleases               4   name mismatch (releases and authorisation
[OK]                RepositoryReleases               4   name mismatch (releases and authorisation
).
```

the 3rd line looks well formatted (cut before the line wrapping), whereas the 4th line is printed as a new line, and includes a line wrapping (for `).`).

CraigFe:
Have reproduced this on `master`. Three years later, what do you suggest we do to fix the formatting in this case? Perhaps not print the `pending` line if it's wider than the terminal?

CraigFe:
On closer inspection, as a result of https://github.com/mirage/alcotest/pull/189 it's no longer simple to fix this issue: we no longer track the width of the terminal.

We could do this without a dependency on `tput` by using the `TIOCGWINSZ` ioctl, as in https://github.com/cryptosense/terminal_size.

PR #282, opened 2020-10-20
Dynamically adapt pretty-printing according to the terminal width

CraigFe:
This PR does two things:

- determines the terminal width dynamically, by adding a dependency on `terminal_size`. (This library currently depends on `Alcotest` for testing, so we may need to vendor it.)

- fixes pretty-printing of long test names, by truncating them with "`...`". This ensures that they never span more than one line, which previously caused trouble for carriage returns (fix https://github.com/mirage/alcotest/issues/72) and the unicode boxes (fix https://github.com/mirage/alcotest/issues/277).


CraigFe:
Looks like `terminal_size` doesn't support Windows, so we'll need to replicate the stubs here (either with a default for Windows or using a [different solution](https://stackoverflow.com/questions/6812224/getting-terminal-size-in-c-for-windows)). @emillon: any thoughts on which approach to take?

CraigFe:
I scrapped the idea of using `terminal_size` and wrote my own set of stubs which can support Windows + don't have the cyclic dependency issue.

This seems to pass CI on all tested platforms, so can be considered for merge.

CraigFe:
Will merge this next week if there are no objections.

src/alcotest/alcotest_stubs.c
```diff
@@ -0,0 +1,82 @@
...
+	CAMLreturn(result);
+}
+
+// Unsupported platform
+#else
+
+CAMLprim value ocaml_alcotest_get_terminal_dimensions(value unit)
+{
+	CAMLparam1(unit);
+	CAMLlocal2(result, pair);
```
emillon, line 76: Not that it changes anything
```suggestion
	CAMLlocal1(result);
```

CraigFe: Indeed, thanks.

PR #297, opened 2021-04-14
Don't print empty assertions

CraigFe:
This ensures that passing assertions with an empty message are silent, which is probably the correct behaviour.


CraigFe:
Travis CI test failure unrelated, so merging.

PR #299, opened 2021-04-17
alcotest.opam: add lower bound on cmdliner

mseri:
The Ocaml CI sometimes fails with `Unbound value Cmdliner.Arg.conv`, which was introduced in version 1.0.0


CraigFe:
Thanks @mseri. I've made this change to `dune-project`, which should fix the CI.

PR #302, opened 2021-05-03
Extract Log_trap abstraction and cleanup log handling

CraigFe:
Pulls out a `Log_trap` component for capturing test logs in files.

This is another chunk of refactoring extracted from https://github.com/mirage/alcotest/pull/294, which requires some more complicated redirection logic.

### DIFF
--- a/src/alcotest-engine/utils.ml
+++ b/src/alcotest-engine/utils.ml
@@ -1,5 +1,9 @@
 let ( >> ) f g x = x |> f |> g
 
+module Fun = struct
+  let id x = x
+end
+
 module String = struct
   include Astring.String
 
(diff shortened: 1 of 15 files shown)


src/alcotest-engine/utils.ml
```diff
@@ -1,5 +1,9 @@
 let ( >> ) f g x = x |> f |> g
 
+module Fun = struct
+  let id x = x
+end
```
Ngoguey42, line 5: What's wrong with the one from `Stdlib`? :)

Ngoguey42 (commented):
LGTM

src/alcotest-engine/utils.ml
CraigFe: It doesn't exist pre-4.08, sadly.

PR #303, opened 2021-05-04
Extract standard library extensions to a package-private lib

CraigFe:
Yet more refactoring. Alcotest v2 needs a lot more extensions to the standard library (type identifiers, equalities, UTF8 strings etc.), so it makes sense to pull all that out to its own library.


Ngoguey42 (commented):
LGTM :)

PR #304, opened 2021-05-06
Use shorter random IDs for test runs

CraigFe:
Currently we use `uuidm` for generating test IDs. These seem a bit unnecessarily long to display / view when debugging: 8 alphanumeric characters is already enough for 3 trillion unique IDs, so I've gone with that.


Ngoguey42 (commented):
LGTM

PR #305, opened 2021-05-06
Change the `~filter` test selector to be a predicate

CraigFe:
The existing type of this argument is very obscure – and was added purely to
keep the CLI in correspondence with the runtime config – and so it's seen no
use in the Opam repository. This changes the type to be more sensible, in a way
that simplifies the V2 PR.


Ngoguey42 (commented):
LGTM

PR #312, opened 2021-05-20
Fix the processing of test filter ranges to match the documentation

CraigFe:
The documentation of the `test` subcommand gives an example of an integer range using the `-` delimiter, when the implementation actually uses `..`. This regression was introduced by cb9cf430a9f2ad152b2e844c462115d3c27adb71, which switched from `-` to `..` internally.

This PR takes the most liberal approach of supporting both syntaxes (and documents them appropriately).

CC @samoht.


CraigFe:
For some context, the switch from `-` to `..` was initially proposed [here](https://github.com/mirage/alcotest/issues/156#issuecomment-516049169) in order to leave room for `-` to mean "negation".

I think this was a good rationale, but in the time since then we haven't ended up implementing that feature. Personally, I find `-` less surprising and so suggest we support that until there's a specific reason not to, but ultimately I mostly care that the docs match the implementation.

src/alcotest-stdlib-ext/alcotest_stdlib_ext.ml
```diff
@@ -45,6 +45,15 @@ module List = struct
...
     aux [] n l
 
+  let concat_map f l =
+    let rec aux f acc = function
+      | [] -> rev acc
+      | x :: l ->
+          let xs = f x in
+          (aux [@tailcall]) f (rev_append xs acc) l
+    in
+    aux f [] l
```
Ngoguey42, line 55: Any reason you are passing `f` to `aux`?

Ngoguey42 (approved):
LGTM

src/alcotest-stdlib-ext/alcotest_stdlib_ext.ml
CraigFe: This is just me over-optimising out of habit. `ocamlopt` w/o flambda doesn't do lambda lifting, so if you don't string `f` through the calls to `aux` then you need to allocate a closure to store it (which is slower).

https://godbolt.org/z/4zxvszW7c

Ngoguey42: Thanks :)

PR #318, opened 2021-08-16
Print strings and chars with OCaml syntax on assertion failures

CraigFe:
In particular, wrap them with single- or double-quotes and add string escapes. This makes the various printers more consistent & ensures that control characters such as `\r` in the test data don't mess with the test output.


Ngoguey42 (approved):

samoht:
LGTM

PR #323, opened 2021-09-27
[RFC] Javascript backends for Alcotest

patricoferris:
Hello! First of all, thank you to the maintainers of Alcotest. I think every piece of OCaml I've ever written uses it, except my JS bindings and code... until now!

Related Issue: #120 

*Note: this is a draft PR just to have a discussion on this topic, I realise it is quite a big dependency to pull in for potentially not a very large user base...*

There's a companion repository that shows a more real-world example binding [js-search](https://github.com/bvaughn/js-search) and testing it in node and in the browser: https://github.com/patricoferris/alcotest-js-example (run the tests in your browser by visiting https://patricoferris.github.io/alcotest-js-example/)

### Backends

There tends to be two backends for javascript. One being [node](https://nodejs.org/en/) and the other is the browser. Both are quite different environments with node being able to provide the same functionality (more or less) as the Unix backend and the browser being more akin to Mirage (no file-system).

This PR provides a virtual package, `alcotest-js`, that has two backends `alcotest-js.node` and `alcotest-js.browser`. Since there's no way to immediately invoke a JS executable in dune (like in a `(test...` stanza with Unix), people using the library will always (a) build the JS using js_of_ocaml and (b) run the tests either in node or the browser.

#### Node

The node example in this PR adds a simple rule to run node (node is the default implementation):

```
(executable
 (name simple)
 (modes js)
 (libraries alcotest-js))

(rule
 (alias nodetest)
 (deps simple.bc.js)
 (action
  (run node %{deps})))
```

Apart from being a little slower than a normal executable, everything should be the same as if you were using the normal Unix backend.

#### Browser

The browser is a little different. The idea here is that you just produce a single-page application of your tests with the logs streamed into the page as they run. I have experimented with tools like [puppeteer](https://github.com/puppeteer/puppeteer) which uses headless chrome to capture the output and promote it, to try and provide a similar experience as Unix/node... it sort of works.

In the demo repository, I simply copy it the `docs` folder so you can run it: https://patricoferris.github.io/alcotest-js-example/. Integrating this into CI would be challenging, but I've already found it quite useful as is. It tests libraries that report to be compatible with node and the browser.

### Known Issues/Questions

 - Logging in the browser back-end isn't quite right, I didn't fully understand the `with_redirect` function but I think the issue is there (perhaps I should stop the flushing to HTML at some point).
 - As I mentioned, it's quite challenging to get the CLI experience of the Unix/Node backends with the browser one. Any ideas here would be great.
 - The node backend has a small `Jv` module which provides a similar API as the Brr `Jv` module, it would be great to re-use that but when running `dune build` which does the separate compilation and no dead-code elimination, browser things were brought into the JS output that `node` was not happy about 🤷  ?
 - The node redirection logic feels a little... hacky?


hhugo:
I've just opened #324 which I believe could replace the node backend here

hhugo:
I've pushed a PoC for browser support based on #324 and reusing some of your logic. see  https://github.com/hhugo/alcotest/commit/https://github.com/hhugo/alcotest/commit/d735a6eb7be10eed908af73f644b40f954ccde9a.
In short, browser support can be implemented with the existing backend by just running additional initialization.
One just need to call the init function bellow.
```ocaml
let init () =
  let open Js_of_ocaml in
  let open Brr in
  let state = Current_ansi.create () in

  let output =
    let out = El.div [ El.h1 [ El.txt' "Alcotest-js" ] ] in
    El.set_at (Jstr.v "style")
      (Some
         (Jstr.v
            "font-family: monospace; color: #e8e8e8; background: #131212; \
             padding: 2em;"))
      out;
    El.(set_prop Prop.id (Jstr.v "output") out);
    let style = El.style [ El.txt' Current_ansi.css ] in
    El.append_children (Document.body G.document) [ style; out ];
    out
  in

  let get_or_make name =
    match Document.find_el_by_id G.document (Jstr.v name) with
    | Some v -> v
    | None ->
       let d = El.div [] in
       El.append_children output [ d ];
       El.(set_prop Prop.id (Jstr.v name) d);
       d
  in
  let append name s =
    let s = Current_ansi.process state s in
    let p = El.pre [] in
    El.to_jv p |> fun jv ->
                  Jv.set jv "innerHTML" (Jv.of_string s);
                  El.append_children (get_or_make name) [ p ]
  in
  Sys_js.set_channel_flusher stdout (fun content -> append "stdout" content);
  Sys_js.set_channel_flusher stderr (fun content -> append "stderr" content);
  ()
```

patricoferris:
@CraigFe thanks for taking a quick look :)) It looks like @hhugo's work is more promising given the recent addition of node support when using unix commands in js_of_ocaml and the latest reply shows there's no need for a browser backend either. 

Re:testing maybe we just need a `conf-node` in opam-repository? Feel free to ping me in #324 if I can help in any way (more testing etc.). For now to keep the conversation all in one place I'll close this draft PR. Again thanks for taking a look :))

hhugo:
It might make sense to package the snippet of code above to ease setup the testing inside the  browser

PR #324, opened 2021-09-28
Make alcotest work with js_of_ocaml

hhugo:
- partially fix https://github.com/mirage/alcotest/issues/120
- could replace the node backend implemented in https://github.com/mirage/alcotest/pull/323

I've also created the same change against the latest version released 1.4.
See https://github.com/hhugo/alcotest/tree/alcotest.1.4+jsoo

This PR requires an unreleased version of js_of_ocaml-compiler https://github.com/ocsigen/js_of_ocaml/pull/1149

### DIFF
--- a/examples/bad/dune
+++ b/examples/bad/dune
@@ -1,5 +1,6 @@
 (executable
  (name bad)
+ (modes native js)
  (libraries alcotest astring))
 
 ; This fails at runtime, so just build the executable
@@ -8,3 +9,8 @@
  (name runtest)
  (package alcotest)
  (deps bad.exe))
+
+(alias
+ (name runtest-js)
+ (package alcotest)
+ (deps bad.bc.js))
--- a/examples/dune
+++ b/examples/dune
@@ -1,4 +1,15 @@
 (tests
  (names simple floats)
  (package alcotest)
+ (modes native js)
  (libraries alcotest))
+
+(rule
+ (alias runtest-js)
+ (action
+  (run node %{dep:floats.bc.js})))
+
+(rule
+ (alias runtest-js)
+ (action
+  (run node %{dep:simple.bc.js})))
--- a/src/alcotest-lwt/dune
+++ b/src/alcotest-lwt/dune
@@ -1,4 +1,4 @@
 (library
  (name alcotest_lwt)
  (public_name alcotest-lwt)
- (libraries logs lwt lwt.unix fmt alcotest.engine alcotest))
+ (libraries logs lwt fmt alcotest.engine alcotest))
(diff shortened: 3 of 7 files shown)


hhugo:
`with_redirect` is implemented using the same mechanism used by ppx_expect (which already work with js_of_ocaml out of the box). see https://github.com/janestreet/ppx_expect/tree/master/collector

CraigFe:
Thanks @hhugo for the contribution. One initial thought: what's the reasoning for copying code from `caml/{io,sys}.h` (and is it possible to do without that)? I'd rather not introduce potential liabilities w.r.t. version compatibility and licensing if possible.

hhugo:
I didn't spend too much time on this. This time, I just duplicated the logic from ppx_expect.
I don't really remember why I did it this way when adding the support inside ppx_expect, I suspect it's just that the code was guarded by CAML_INTERNALS.

hhugo:
On testing, do you have any interest switching to ocaml-setup for the CI ? @smorimoto  would you be willing to help ?

hhugo:
The duplicated code is indeed not exported (guarded by CAML_INTERNALS).
I still don't have a better approach to propose, sadly.

hhugo:
I've managed to reduce the code duplicated from ocaml to 3 lines only.

hhugo:
@CraigFe, I'm planning to make a new release of jsoo in couple of days. We can merge this PR after. What do you want to do for the CI ? I can help you if you switch to ocaml-setup (installing node in couple of lines https://github.com/ocsigen/js_of_ocaml/blob/master/.github/workflows/build.yml#L51) but I can't help with ocaml-ci I think.

smorimoto:
@hhugo I'm happy to help you set up. The problem with the workflow not being triggered is here:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
and it means that only in this PR, the Mirage team will have to manually approve each time.

CraigFe:
Thanks @hhugo for minimising the code duplication.

Regarding CI, I'm happy to pick up a GitHub actions workflow for running the Node CI if that's the easiest route. We already use both GH actions and AppVeyor to cover gaps in what `ocaml-ci` offers, so there's precedent for that. (I don't want to replace `ocaml-ci` completely, as its coverage / speed is hard to beat.)

I'll review this weekend, and then we can merge & cut an `alcotest.1.5.0` with Jsoo support.

hhugo:
Would you like to replace appveyor  by GitHub action while we are at it ?

CraigFe:
Sounds good to me :-)

hhugo:
I'm experimenting with the ci in https://github.com/hhugo/alcotest/pull/2

hhugo:
closing in favor of https://github.com/mirage/alcotest/pull/326

PR #325, opened 2021-09-28
Fix bindings to `TIOCGWINSZ` on MacOS

CraigFe:
... by using bindings from the `terminal` library instead, as these handle the MacOS case correctly.

See https://github.com/CraigFe/progress/blob/5a221d16bb068ac8a500ee90644055ffbabba034/src/terminal/terminal_stubs.c.


arbipher:
It works perfectly on my mac now. Thank you.

PR #335, opened 2021-12-18
Fix compilation on bytecode architectures

glondu:
Patch applied in Debian, so that alcotest's build and tests pass on all Debian architectures.


samoht:
I've rebased and force-puhsed - Thanks for the patch, it'll make it into the next release

samoht (approved):

PR #342, opened 2022-02-25
Correctly lower-bound on OCaml 4.05

MisterDA:
`alcotest >= 1.5.0 → fmt >= 0.8.7 → ocaml >= 4.05.0`
Also set OCamlFormat parser to OCaml 4.05 to make sure it doesn't reformat using newer syntax.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,5 +1,6 @@
-version = 0.19.0
+version = 0.20.1
 profile = conventional
+ocaml-version = 4.05.0
 
 module-item-spacing = compact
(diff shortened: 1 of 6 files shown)


CraigFe (approved):

CraigFe:
Thanks. I think it's debatable whether this is the "correct" lower-bound for us to be stating for Alcotest, since the lower-bound implied by transitive dependencies isn't known in general, but I'm happy to set 4.05.0 as our limit for now (potentially bumping to 4.06 before the next release).

PR #348, opened 2022-03-02
Make alcotest-lwt compatible with jsoo

hhugo:


hhugo:
gentle ping

CraigFe:
Thanks @hhugo. I'll review this (& go through backlog elsewhere) this weekend :+1:

src/alcotest-lwt/dune
```diff
@@ -7,5 +7,4 @@
   fmt
   alcotest.engine
   alcotest
-  (re_export lwt.unix) ;; Unused dependency added for convenience to consumers
```
TheLortex, line 10: This looks like a breaking change.. I don't know how many packages depend on lwt.unix being re-exported though.

test/e2e/alcotest-lwt/failing/async_failure.ml
```diff
@@ -5,11 +5,11 @@ let free () =
 let test_lwt switch () =
   Lwt_switch.add_hook (Some switch) free;
   Lwt.async (fun () -> failwith "All is broken");
-  Lwt_unix.sleep 10.
```
TheLortex, line 8: What was the reason for this 10 seconds sleep ?

test/e2e/helper/alcotest_test_helper.ml
```diff
@@ -0,0 +1,7 @@
+let rec wakeup_until_resolved p =
```
TheLortex, line 1: It should be explained somewhere that to make the tests work with js we use a custom lwt event loop. 
Either in [CONTRIBUTING.md](https://github.com/mirage/alcotest/blob/main/CONTRIBUTING.md#expect-tests) or in `alcotest_test_helper.ml`.

TheLortex (changes requested):
Thanks for this PR. 

The changes are making sense to me. However I'm afraid of the breaking changed introduced by the removal of the `lwt.unix` dependency. I'll try to investigate and see which packages are involved. 

I have added a comment about `alcotest_test_helper`, we should explain somewhere why it is used instead of `Lwt_main.run`.

Can you rebase on top of `main` and add a changelog entry ?

MisterDA:
@hhugo what's the status of this?

hhugo:
I don't have time to spend on this and no longer have a need for this at the moment. Feel free to close or takeover the PR

PR #349, opened 2022-03-26
Generalize the type of check_raises

favonia:
Rationale: (1) the function we wish to test might not have `unit` as its codomain type, and (2) the dropping is not silent; if some value was returned and then dropped, the check would fail and the programmer would know.

Question: I did not touch `CHANGES.md` because I could not find a section for unreleased features. Maybe I missed something?


favonia:
Closing this due to lack of interest.

PR #339, opened 2022-02-14
Update to cmdliner.1.1.0

MisterDA:

PR #350, opened 2022-04-14
refactor: update to cmdliner 1.1.0

rgrinberg:
Would be nice to make a point release with this as it would make it possible for me to submit an updated alcotest without disabling tests.


rgrinberg:
Looks like this is done in #339

PR #351, opened 2022-04-16
Add missing dependency on Unix

dra27:
Silences the alert proposed in ocaml/ocaml#11198, but this PR can be merged regardless.

### DIFF
--- a/src/alcotest/dune
+++ b/src/alcotest/dune
@@ -1,6 +1,6 @@
 (library
  (public_name alcotest)
- (libraries alcotest.engine astring fmt fmt.tty)
+ (libraries alcotest.engine astring fmt fmt.tty unix)
  (foreign_stubs
   (language c)
   (names alcotest_stubs))


hannesm:
I'm not sure why this is needed (I don't see a dependency to "Unix" in src/alcotest-engine)... especially in #238 there was quite some work to remove the Unix dependency to allow alcotest to execute on MirageOS.

dra27:
I'm not sure why I'd added the dependency to alcotest-engine - I would guess that in the myriad packages I was patching, I read a functor application and then applied it to the wrong package!

PR updated - the alcotest library definitely uses `Unix` (this clears the warning on 5.x)

dinosaure:
In anyway, `fmt.tty` uses `Unix` so transitively, `alcotest` uses `Unix` too.

dra27:
Except that the project has `(implicit_transitive_deps false)`, so it can’t use Unix without declaring the dependency - it just worked before because it was in the stdlib

TheLortex (approved):

PR #360, opened 2022-10-20
Fix warnings triggered by Dune 3

MisterDA:


src/alcotest-engine/pp_intf.ml
```diff
@@ -71,10 +71,10 @@ module type S = sig
 
   val with_surrounding_box : 'a Fmt.t -> 'a Fmt.t
   (** Wraps a formatter with a Unicode box with width given by
-      {!X.stdout_columns}. Uses box-drawing characters from code page 437. *)
+      {!_.stdout_columns}. Uses box-drawing characters from code page 437. *)
```
TheLortex, line 74: What is `_.stdout_columns` referring to ?

src/alcotest-engine/pp_intf.ml
```diff
@@ -71,10 +71,10 @@ module type S = sig
...
 
   val horizontal_rule : _ Fmt.t
-  (** Horizontal rule of length {!X.stdout_columns}. Uses box-drawing characters
+  (** Horizontal rule of length {!_.stdout_columns}. Uses box-drawing characters
```
TheLortex, line 77: Same

TheLortex (approved):
LG™ 

One minor comment about the documentation

src/alcotest-engine/pp_intf.ml
MisterDA: It's odd, but the `X` refers to the argument of the functor `Make` below.
`module Make (X : Make_arg) : S`
Dune wants me to remove the `X` name and replace it with `_` since the binding is not used. Odoc doesn't let me refer to `Make_args.stdout_columns`, so I use `_` which is the new name of the functor argument, and odoc and dune are happy and generate correct doc.

TheLortex: I see, thanks for the clarification.

PR #363, opened 2022-11-17
Minor patches: consistent naming and allocation in C stubs

MisterDA:
(I dont think this requires a changelog entry)


samoht:
Thanks

PR #364, opened 2022-11-17
Basic CI integration

MisterDA:
This PR attempts to do simple integration from Alcotest to CI systems. For now, Alcotest detects the CI system using the `CI`, `GITHUB_ACTIONS`, or `OCAMLCI` environment variables. As a first showcase of what we can do, when detecting GitHub Actions it will output annotations allowing to fold/unfold a test suite in the web interface. When detecting it's running in CI, Alcotest will also enable the show_errors modes (this can be overridden via the API) so that all failing tests are displayed in the CI (otherwise users would have to fix the failing test, push, and so on, to see the other errors).
I plan to expand the features if this PR is well received.

### DIFF
--- a/src/alcotest-engine/config_intf.ml
+++ b/src/alcotest-engine/config_intf.ml
@@ -2,6 +2,9 @@ module Types = struct
   type bound = [ `Unlimited | `Limit of int ]
   type filter = name:string -> index:int -> [ `Run | `Skip ]
 
+  type ci = [ `Github_actions | `OCamlci | `Unknown | `Disabled ]
+  (** All supported Continuous Integration (CI) systems. *)
+
   type t =
     < and_exit : bool
     ; verbose : bool
@@ -13,7 +16,8 @@ module Types = struct
     ; filter : filter option
     ; log_dir : string
     ; bail : bool
-    ; record_backtrace : bool >
+    ; record_backtrace : bool
+    ; ci : ci >
 
   type 'a with_options =
     ?and_exit:bool ->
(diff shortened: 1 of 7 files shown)


src/alcotest-engine/config_intf.ml
```diff
@@ -13,7 +13,8 @@ module Types = struct
     ; filter : filter option
     ; log_dir : string
     ; bail : bool
-    ; record_backtrace : bool >
+    ; record_backtrace : bool
+    ; ci : [ `Github_actions | `OCamlci | `Unknown | `Disabled ] >
```
samoht, line 17: Could you declare this as a separate type?

samoht:
I think you forgot to promote the expect test outputs - might be worth being explicit if these environment variables are set in the tests (so we can also check these are working as expected).

src/alcotest-engine/config_intf.ml
MisterDA: Done, thanks!

MisterDA:
> I think you forgot to promote the expect test outputs - might be worth being explicit if these environment variables are set in the tests (so we can also check these are working as expected).

I've set `CI=false` in the environment of the tests. Testing the CI will likely need to duplicate the tests. As the features are quite simple for now, maybe that can be done later?

MisterDA:
For reference, this is the language GitHub Actions uses for annotations: [Workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions).
Here's an open feature request of mine: [Grouping log lines: set the folded/unfolded state with ::endgroup:: at the end of the logs](https://github.com/community/community/discussions/39411).

samoht:
Many thanks!

PR #366, opened 2022-11-21
Best-effort automatic reporting of the location of `Alcotest.check`

MisterDA:
This is an attempt of automatically discover the location of calls to `Alcotest.check`, in order to not force the user to pass `~here` or `~pos` arguments. It uses `Printexc.get_callstack`. It's an early attempt that is fragile and requires OCaml 4.11.
The test need to end with the unit value in order to prevent the compiler from optimizing the last call to a tail-call, erasing the location in the process. Is there a way to change Alcotest' code so that this optimization doesn't happen?
A more advanced attempt should probably walk the call stack to find recognizable enclosing functions calls, add regression testing, and make it work (or default to no-op) on pre-4.11. I don't know whether the code is efficient enough.

Try it out with
```shell
dune exec --display=quiet -- examples/bad/bad.exe -e
```

cc @TheLortex

### DIFF
--- a/src/alcotest-engine/test.ml
+++ b/src/alcotest-engine/test.ml
@@ -192,6 +192,10 @@ let check (type a) ?here ?pos (t : a testable) msg (expected : a) (actual : a) =
       ()
     and pp_actual ppf () =
       Fmt.pf ppf "   Received: `%a'" (styled `Red (pp t)) actual
+    and here, pos =
+      match (here, pos) with
+      | None, None -> (Callsite_loc.get (), None)
+      | _ -> (here, pos)
     in
     raise
       (Core.Check_error
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ### dev
 
+- Try automatically reporting the location of calls to Alcotest.check.
+  (#366, @MisterDA, review by @TheLortex)
+
 - Detect that Alcotest is running in CI and change output accordingly.
   (#364, @MisterDA)
 
--- a/src/alcotest-engine/callsite_loc.412.ml
+++ b/src/alcotest-engine/callsite_loc.412.ml
@@ -57,3 +57,12 @@ let get ?(__FUNCTION__ = "Alcotest_engine__Test.check") () =
         | None -> None)
     | _ -> None
   else None
+
+let get ?__FUNCTION__ () =
+  let guess =
+    match Sys.getenv "ALCOTEST_SOURCE_CODE_POSITION" with
+    | "" | "false" | "no" | "n" | "0" -> false
+    | "true" | "yes" | "y" | "1" -> true
+    | _ | (exception _) -> true
+  in
+  if guess then get ?__FUNCTION__ () else None
--- a/src/alcotest-engine/callsite_loc.412.ml
+++ b/src/alcotest-engine/callsite_loc.412.ml
@@ -1,0 +1,59 @@
+(*
+ * Copyright (c) 2022 Antonin Décimo <antonin@tarides.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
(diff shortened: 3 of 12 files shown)


src/alcotest-engine/test.ml
```diff
@@ -177,6 +177,41 @@ let pp_location =
...
+        let len = raw_backtrace_length raw_backtrace in
+        if len >= 3 then
+          let slot0 =
+            get_raw_backtrace_slot raw_backtrace 0 |> convert_raw_backtrace_slot
+          and slot2 =
+            get_raw_backtrace_slot raw_backtrace 2 |> convert_raw_backtrace_slot
+          in
+          match (Slot.name slot0, Slot.name slot2) with
+          | ( Some "Alcotest_engine__Test.check",
+              Some "Alcotest_engine__Core.Make.protect_test.(fun)" ) -> (
```
TheLortex, line 197: One problem I see here might be that there can be multiple call layers between `Core.Make.protect_test` and `Test.check`. 
For example if the user's test calls into an intermediate function that calls into `check`, then `protect_test` would be present in slot `3` instead of `2`.

Maybe it's better do that the other way around: check that slot `0` is `check` and that `1` is _not_ `protect_test` ?

MisterDA: You're right! I've changed that.

MisterDA:
Reworked the code a bit thanks to @TheLortex review.
It now supports pre-4.11, is a bit more robust, and includes tests.
This only works with `Alcotest.check` and `Alcotest.check'`, not with `Alcotest.fail` and related functions, because they're always in tail-call position since they raise, and adding a statement after these is unsound…
There's no way to force the compiler not to optimize the tail-call of `Alcotest.check`, so I've added a note explaining how Alcotest may "guess" the location.

TheLortex:
Thanks for this PR. It's a _free_ improvement of the location reporting feature of alcotest. 

The drawback is that it's not simple to predict when it works (backtrace enabled and call to `Alcotest.check` is not a tailcall). 
The addition of a _return unit_ instruction after the check to eliminate the tailcall might be somewhat brittle (maybe it will be optimized away at some point) and it's hard to discover its reason without having the context (meaning the unsuspecting user will want to remove it) 
```diff
- Alcotest.check ... 
+ Alcotest.check ... ;
+ ()
```

I wonder what other alcotest users think about it.

MisterDA:
I've changed the code to use the latest `Printexc` API, introduced in 4.12. Alcotest will fallback to a no-op before that.
I've also added an environment variable to disable position guessing: for instance Alcotest uses expect tests on its output, and the tests would fail on earlier OCaml versions. I don't expect they'll be many users of it apart from Alcotest itself.
I've added regression tests. Maybe they won't really work and `strip_randomness` should be modified to remove that extra information instead?

samoht:
Thanks!

Could you show an example of what this changes in the output?

Also, for the new files, could you keep the same license as the rest of the project?

MisterDA:
This just shows the position of the call to `Alcotest.check` in the test source file, if available. The position is almost identical to using `Alcotest.check ~here:[%here]` from ppx here or `Alcotest.check ~pos:Stdlib.__POSITION__` but the user don't have to specify the location themselves.
It only works if the call to `Alcotest.check` is a tailcall (usually the last one is not), and backtraces have been enabled (usually the case); and with OCaml >= 4.12.

An output with the location reported:
```
ASSERT strings
File "test/e2e/alcotest/source_code_position/test_source_code_position.ml", line 37, character 2:
FAIL strings

   Expected: `"A"'
   Received: `"B"'
```

one with disabled:
```
ASSERT strings
FAIL strings

   Expected: `"A"'
   Received: `"B"'
```

> Also, for the new files, could you keep the same license as the rest of the project?

Done.

samoht:
That looks like a great feature!

```
ASSERT strings
--
507 | -File "test/e2e/alcotest/source_code_position/test_source_code_position.ml", line 37, character 2:
508 | +File "test/e2e/alcotest/source_code_position/test_source_code_position.ml", line 24, character 2:
509 | FAIL strings
```
However, it doesn't seem very stable across OCaml versions 😁 Do we care? Should we just disable it for older OCaml versions?

samoht:
Looks great, thanks for the patch!

PR #368, opened 2022-12-13
Allow skipping a test case from inside the test case

apeschar:
This is useful when you want to perform logic to check whether or not some test case can run in the current environment.

~~~ocaml
let test_only_on_windows () =
  if Sys.os_type <> "Win32" then Alcotest.skip () else ();
  (* actual test code *)
~~~

### DIFF
--- a/src/alcotest-engine/core.ml
+++ b/src/alcotest-engine/core.ml
@@ -19,7 +19,6 @@ open! Import
 open Model
 
 exception Check_error of unit Fmt.t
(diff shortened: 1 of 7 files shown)


samoht:
Hi! Thanks for your contribution, that looks like a good idea!

Out of curiosity, in what context do you use Alcotest? Are you building interesting projects with it? I'm always interested to hear about new use-cases :-)

PR #370, opened 2023-02-04
syntax highlighting: shell → sh-session

toastal:
These blocks are not shell scripts, but rather the shell session in a terminal. This syntax highlighting change reflects that.

To see it rendered as a preview, check the branch @ https://github.com/toastal/alcotest/blob/sh-session/README.md


TheLortex (approved):
Thanks for this change !

PR #374, opened 2023-03-01
opam, doc, backtraces w/ bytecode fixes

MisterDA:


samoht (approved):

MisterDA:
A bit complicated, but I managed to get the test with the locations extracted from the backtraces to always run in bytecode (even when native compilation is available), and correctly on bytecode-only architectures. No need to disable it anymore!
Sadly, I haven't found a way to conditionally enable dune rules based on whether native compilation is available.

Tested on Linux x86_64, i386, and Windows x86_64.

Issue #322, opened 2021-09-20
Manually set output width


arbipher:
Is it possible to set output width manually?

I searched the issues and found this [c_stub](https://github.com/mirage/alcotest/blob/master/src/alcotest/alcotest_stubs.c#L53), but I haven't figured out how to set it, or how I can set `dimension`.

It will be convenient for people who have test cases with long names. Currently with the `...` it's hard to know which test cases are failed.

aslanix:
Seconding this. Forcing truncating of the lines is a bit of an anti-pattern.

CraigFe:
I agree that truncation shouldn't be the default in cases where Alcotest can't successfully determine the terminal width. The fix  would be to avoid setting a default [here](https://github.com/mirage/alcotest/blob/2305f78331d1675a62f74920e2bce2380e14c597/src/alcotest-engine/pp.ml#L51), and them plumb that through to the other pretty-printers in the file (which should then not truncate in the `None` case). While we're at it, we could consider setting the margin on the `Format.formatter` to match the one that's actually detected.

Regarding manually overriding the determined width, it's already possible (but annoying) to do this by building a different `Platform` to sit underneath the Alcotest runner:

```ocaml
let () =
  let module Platform (M : Alcotest_engine.Monad.S) = struct
    include Alcotest.Unix (M)

    let stdout_columns () =
      match stdout_columns () with None -> Some 200 | Some _ as x -> x
  end in
  let module Runner = Alcotest_engine.Cli.Make (Platform) (Alcotest_engine.Monad.Identity) in
  Runner.run __FILE__
    [
      ( "foo",
        [
          Alcotest.test_case
            "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec \
             urna nisi, mattis scelerisque nulla at, rutrum convallis ante. \
             Vestibulum bibendum ex at purus gravida, ac pellentesque elit \
             facilisis. Etiam lacinia eros fermentum lacinia ornare. \
             Vestibulum facilisis mollis dolor ultrices lobortis. In a nisl id \
             purus rutrum accumsan non ac ligula."
            `Quick (fun () -> assert false);
        ] );
    ]
```

[...]

PR #381, opened 2023-03-27
Fix reporting terminal dimensions

### LINKED ISSUE #356: Random crashes when running tests in a non-standard terminal
I sometimes execute our test written with Alcotest in Emacs. Recently I have seen them to crash with the following exception:
```
main.exe: internal error, uncaught exception:
          Division_by_zero
          Raised at Alcotest_engine__Pp.Make.event in file "src/alcotest-engine/pp.ml", line 165, characters 11-59
          Called from Alcotest_engine__Core.Make.perform_test in file "src/alcotest-engine/core.ml", line 237, characters 4-45
          Called from Alcotest_engine__Monad.Extend.Syntax.(>|=).(fun) in file "src/alcotest-engine/monad.ml", line 32, characters 46-51
          Called from Alcotest_engine__Monad.Extend.List.fold_map_s.inner in file "src/alcotest-engine/monad.ml", line 44, characters 26-34
          Called from Alcotest_engine__Core.Make.perform_tests in file "src/alcotest-engine/core.ml", line 252, characters 6-276
          Called from Alcotest_engine__Core.Make.result in file "src/alcotest-engine/core.ml", line 302, characters 19-44
          Called from Alcotest_engine__Core.Make.run_tests in file "src/alcotest-engine/core.ml", line 356, characters 8-27
          Called from Alcotest_engine__Core.Make.run_with_args' in file "src/alcotest-engine/core.ml", line 390, characters 6-304
          Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 24, characters 19-24
          Called from Cmdliner_eval.run_parser in file "cmdliner_eval.ml", line 34, characters 37-44
```
It appears that the `0` comes from the call to `ioctl(STDOUT_FILENO, TIOCGWINSZ, &ws)`, which Emacs apparently sometimes reports badly. It would be nice to make Alcotest resistant to this kind of error and perhaps assume some standard value for the width of the window.

MisterDA:
- Fix division by zero when size of the terminal is incorrectly reported as zero. (fix #356, #381, @MisterDA)

- Enable terminal size reporting on macOS and Windows. Also report the terminal size even when the test is run buffered by Dune.
  (#381, @MisterDA)

- Allow overriding the number of columns with `ALCOTEST_COLUMNS` env var. (#322, #381, @MisterDA)
  
My intent was to use a bigger default than 80 columns when the CI is detected. Sadly, this would seem to require too big of a refactoring to communicate the default size to the `Pp` module, which is also why I haven't made the number of columns configurable programmatically, and exposed it only via an environment variable to fix reproducibility issues for the Unix platform.
Oddly enough, communicating the bounds to the `Format` module with `Format.set_margin` hides the first `[OK]`, `[FAIL]`, etc tag. I've haven't dug why.
I've tested that this on macOS, Windows FreeBSD, and Linux.

### DIFF
--- a/src/alcotest/alcotest_stubs.c
+++ b/src/alcotest/alcotest_stubs.c
@@ -55,13 +55,20 @@ CAMLprim value ocaml_alcotest_get_terminal_dimensions(value unit)
 #elif defined (__unix__) || (defined (__APPLE__) && defined (__MACH__))
 #include <unistd.h>
 #include <sys/ioctl.h>
+#include <fcntl.h>
 
 CAMLprim value ocaml_alcotest_get_terminal_dimensions(value unit)
 {
 	CAMLparam1(unit);
 	CAMLlocal2(result, pair);
 	struct winsize ws;
-	int z = ioctl(STDOUT_FILENO, TIOCGWINSZ, &ws);
+	int fd = open("/dev/tty", O_RDONLY | O_NONBLOCK | O_CLOEXEC);
+	if (fd < 0) {
+		result = Val_none;
+		CAMLreturn(result);
+	}
+	int z = ioctl(fd, TIOCGWINSZ, &ws);
+        close(fd);
 	if (z == 0)
 	{
 		pair = caml_alloc_tuple(2);
--- a/src/alcotest/alcotest_stubs.c
+++ b/src/alcotest/alcotest_stubs.c
@@ -69,7 +69,7 @@ CAMLprim value ocaml_alcotest_get_terminal_dimensions(value unit)
 	}
 	int z = ioctl(fd, TIOCGWINSZ, &ws);
         close(fd);
-	if (z == 0)
+	if (z != -1)
 	{
 		pair = caml_alloc_tuple(2);
 		Store_field(pair, 0, Val_int(ws.ws_row));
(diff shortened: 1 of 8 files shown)


MisterDA:
@OlivierNicole has done a nice review for this PR, thanks.
I think it can be merged as-is. Does a anyone have an opinion?

src/alcotest/alcotest_stubs.c
```diff
@@ -56,17 +52,24 @@ CAMLprim value ocaml_alcotest_get_terminal_dimensions(value unit)
...
+#include <fcntl.h>
 
 CAMLprim value ocaml_alcotest_get_terminal_dimensions(value unit)
 {
 	CAMLparam1(unit);
 	CAMLlocal2(result, pair);
 	struct winsize ws;
-	int z = ioctl(STDOUT_FILENO, TIOCGWINSZ, &ws);
-	if (z == 0)
+	int fd = open("/dev/tty", O_RDONLY | O_NONBLOCK | O_CLOEXEC);
```
art-w, line 65: In case someone else is looking at this and doesn't know why `STDOUT_FILENO` was replaced by `"/dev/tty"` (like I was), the difference is when the output of alcotest is piped (say through a grep): stdout wouldn't have a size so the formatting would be different (and probably be confusing to users?), while the tty always return the same size (... I don't have strong opinion regarding this change, in general it's weird for piped programs to respect terminal size but it seems fine for alcotest)

art-w (approved):

dinosaure (approved):
LGTM even if I don't really like the allocation on the C side (return `-1` is a possibility for the `None` case but it's not related to this PR).

PR #384, opened 2023-03-31
Set min version of cmdliner to 1.1.1

gridbugs:
cmdliner.1.1.1 removed trailing whitespace from help messages. Building this with cmdliner < 1.1.1 causes the comparison with alcotest-help.txt to fail due to whitespace differences.


MisterDA (approved):
Thanks!

PR #393, opened 2023-09-18
Add new `Alcotest.suite` function to make writing tests easier

### LINKED ISSUE #126: Review how tests are registered

yawaramin:
Fixes #126.

### DIFF
--- a/examples/floats.ml
+++ b/examples/floats.ml
@@ -27,28 +27,31 @@ For more information, please refer to <http://unlicense.org/>
 
 let e = epsilon_float
 
-let nan () =
-  Alcotest.(check @@ float e) "NaN is NaN" nan nan;
-  Alcotest.(check @@ neg @@ float e) "NaN is not number" nan 7.;
-  Alcotest.(check @@ neg @@ float e) "number is not NaN" 8. nan
-
-let infinity () =
-  Alcotest.(check @@ float e) "+∞ is +∞" infinity infinity;
-  Alcotest.(check @@ float e) "-∞ is -∞" neg_infinity neg_infinity;
-  Alcotest.(check @@ neg @@ float e) "+∞ is not -∞" infinity neg_infinity;
-  Alcotest.(check @@ neg @@ float e) "-∞ is not +∞" neg_infinity infinity;
-  Alcotest.(check @@ neg @@ float e) "+∞ is not 3" infinity 3.
-
-let others () =
-  Alcotest.(check @@ float e) "0 is 0" 0. 0.;
-  Alcotest.(check @@ float e) "0 is epsilon" 0. e;
-  Alcotest.(check @@ neg @@ float e) "0 is not 1" 0. 1.;
-  Alcotest.(check @@ neg @@ float e) "1 is not 0" 1. 0.;
-  Alcotest.(check @@ float e) ".3 is .3" (0.1 +. 0.2) 0.3
-
-let edge_set = [ ("NaN", `Quick, nan); ("∞", `Quick, infinity) ]
-let others_set = [ ("others", `Quick, others) ]
-
 let () =
-  Alcotest.run "Float tests"
-    [ ("Edge cases", edge_set); ("Other floats", others_set) ]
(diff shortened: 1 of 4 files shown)


yawaramin:
```
$ make test
dune build @runtest
File "alcotest-help.txt", line 1, characters 0-0:
diff --git a/_build/default/alcotest-help.txt b/_build/default/alcotest-help.txt.actual
index 02d983c..a684448 100644
--- a/_build/default/alcotest-help.txt
+++ b/_build/default/alcotest-help.txt.actual
@@ -1,109 +1,10 @@
-NAME
-       test.exe - Run all the tests.
+Testing `Utils'.
+This run has ID `G98YX3C9'.
 
-SYNOPSIS
-       test.exe [COMMAND] …
-
```

Obviously, I didn't touch this file...

src/alcotest-engine/cli_intf.ml
```diff
@@ -41,6 +41,38 @@ module V1_types = struct
...
+      with_options
+    (** [suite n register] runs the test suite registered by the registration
+        function [register]. [n] is the name of the tested library.
+
+        Other parameters are the same as they are in [run].
+
+        Example usage:
+
+        {[
+        let () =
```
yawaramin, line 55: The ocamlformat settings don't like the formatting of this example. IMHO, the format settings are incorrect :-)

samoht:
Thanks - two remarks:

- I really like using functions in the API instead of exposing concrete types. It's probably too late to make those types abstract as it will break every existing code, but let's start by exposing proper constructors first and see how adoption goes.
- I am not convinced by inlining all the test cases in "heavy" functions. Would you mind doing the minimum possible changes in the examples to keep the test functions separate (when they are big enough like they are right now) and avoid the heavyweight `begin fun () -> ... end` forms when possible?

yawaramin:
Hi @samoht thanks for the review. Regarding your points,

- Agreed
- Do you mean like e.g.:

```ocaml
let string_case_lower_case () =
  Alcotest.(check string) "same string" "hello!" (To_test.lowercase "hELLO!")

let string_case_capitalization () =
  Alcotest.(check string) "same string" "World." (To_test.capitalize "world.")

let string_concat_string_mashing () =
  Alcotest.(check string) "same string" "foobar" (To_test.str_concat ["foo"; "bar"])

let list_concat_list_mashing () =
  Alcotest.(check (list int)) "same lists" [1; 2; 3] (To_test.list_concat [1] [2; 3])

let () =
  Alcotest.suite "Utils" begin fun group ->
    group "string-case" begin fun case ->
      case "Lower case" string_case_lower_case;
      case "Capitalization" string_case_capitalization;
    end;

    group "string-concat" begin fun case ->
      case "String mashing" string_concat_string_mashing;
    end;

    group "list-concat" begin fun case ->
      case ~speed:`Slow "List mashing" list_concat_list_mashing;
    end;
  end
```

Wouldn't this bring us back to the current situation where the test implementations and registrations are separate and can get out of sync? E.g. if I don't have an empty `.mli` and delete one of the cases, I could forget to delete its implementation.

lessp:
Love this! Can we expect this to go in?

samoht:
I'm looking at this again - I think there are 2 things in that patch that needs to be reviewed separately

1.  adding top-level types and function for group of tests. This is uncontroversial a good change and I'll be happy to merge this (given with find the right name, for instance `group_testlist` is a bit weird :p)
2. changing the default tests are registered -- I'm a bit warry of this one without more user feedback. I'm ok to suggest a new way to do things but I would like to avoid breaking existing tests. Once users start to adopt this new way (and give us some feedback) we could consider changing the way we write examples, etc. 

Do you think that makes sense?

samoht:
In particular, I'd like to make sure we take the feedback from #294 into account

PR #394, opened 2023-09-19
Recognize opam-repo-ci

MisterDA:
Recognize Alcotest is running in opam-repo-ci if the `OPAM_REPO_CI` env var is set to `"true"`, and display all the tests (failures) by default.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,8 @@
 
 - Add `match_raises`, a generalized version of `check_raises`
   (#88, #386, @JoanThibault)
-- Update JaneStreet core and async to v0.16 (#390 @tmcgilchrist)  
+- Update JaneStreet core and async to v0.16 (#390 @tmcgilchrist)
+- Recognize opam-repo-ci. (#394, @MisterDA)
 
 ### 1.7.0 (2023-02-24)
 
--- a/alcotest-help.txt
+++ b/alcotest-help.txt
@@ -107,3 +107,7 @@ ENVIRONMENT
            Whether Alcotest is running in OCaml-CI, if set to 'true'. Display
            tests errors.
 
+       OPAM_REPO_CI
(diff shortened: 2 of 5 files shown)


dinosaure:
I'm not really a fan of this PR since it implies that basically alcotest and `opam-repo-ci` are linked. Isn't there another way (on the `opam-repo-ci` side) to display the tests?

MisterDA:
Well, they all define the `CI` env var to `"true"`, so we can use just that.

There's the bonus that GitHub Actions parses the output log for [_workflow commands_](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions), such as 
```
      - name: Create annotation for build error
        run: echo "::error file=app.js,line=1::Missing semicolon"
```
Alcotest recognizes that it's run in GHA and is able to emit these commands to better integrate in the GHA interface, by highlighting errors, or grouping test cases in drop down buttons. We could even have the GitHub interface link to the file where the error is raised, but that's a bit difficult to do considering the structure of Alcotest.

Neither OCaml-CI nor opam-repo-ci currently support such features.

dinosaure:
Hmmmhmm, I still have a doubt about the validity of having to change alcotest according to CI (and not vice versa) - as you mention, the CI environment variable would be sufficient for all CIs. What I have a problem with is the maintainance involved in such a design where the maintaineurs must always be synchronised with these CIs.

We could imagine extensions such as `alcotest-github-actions` or `alcotest-ocaml-ci` (to disociate the core release cycle from these). As it stands, I'm not going to ask you to do more than this PR but the question should remain open: I'm going to make an issue.

yawaramin:
Don't all the CIs define the `CI` env var?

MisterDA:
Closing, I'll switch to detecting `CI` only, since there's no special integration with opam-repo-ci.

Issue #398, opened 2023-10-11
Be able to specify our own formatter `ppf`


dinosaure:
Currently, `alcotest` uses `Format.err_formatter` which can be used by someone else. In the context of parallel computations, a data-race can exist where `alcotest` tries to print something and someone else try to do so too: they use both the same `Format.formatter`. To avoid such situation, it will be nice to specify our own `Format.formatter` when we configure `alcotest`.

PR #399, opened 2023-10-13
Be able to allocate a formatter for stdout/stderr

dinosaure:
This pull-request, as a draft, wants to solve the issue #398 and solve a data-race on `formatter`s used by Alcotest where they are also used by something else like `Logs` in a parallel task. Currently, this PR is a draft because I'm not really sure about the design. Three points are importants:
1) to be coherent into our usage of `stdout`/`stderr` and what is required by `alcotest-engine` (eg. `Platform`), the user is only able to create these values via `Alcotest.Global.make_{stdout,stderr}` to ensure that these values use `Stdlib.std{out,err}`. Indeed, we don't want to let the user to create a `Format.formatter` which does not uses `Stdlib.std{out,err}` (something like `Format.str_formatter` for instance).
2) The library was tested with TSan and [`httpcats`](https://github.com/robur-coop/httpcats) which launches several tasks in parallel and uses `Logs` in these parallel tasks with `Format.{std,err}_formatter`
3) A controversial change was made about `pp_summary` where the newline does not appear anymore but I don't really understand why. Otherwise, the ouput _seems_ the same as before.


src/alcotest-engine/pp.ml
```diff
@@ -212,7 +212,7 @@ struct
       | 0 -> green_s ppf "Test Successful"
       | n -> red ppf "%d failure%a!" n pp_plural n
     in
-    Fmt.pf ppf "%a in %.3fs. %d test%a run.@," pp_failures r.failures r.time
+    Fmt.pf ppf "%a in %.3fs. %d test%a run.@\n" pp_failures r.failures r.time
```
dinosaure, line 215: This is where we have my controversial change about a new line which does not appear anymore. Not sure if it's a story about `Format`'s boxes or the default margin of a `Format.formatter`.

dinosaure:
It will be really pleasant if someone can look on this PR and merge it. Tests are passing and a real issue exists in the actual state of Alcotest.

MisterDA (approved):
LGTM.

PR #403, opened 2024-04-18
Adapt jsoo runtime to new jsoo version

hhugo:
see https://github.com/ocsigen/js_of_ocaml/pull/1602

### DIFF
--- a/alcotest.opam
+++ b/alcotest.opam
@@ -32,6 +32,7 @@ depends: [
 ]
 conflicts: [
   "result" {< "1.5"}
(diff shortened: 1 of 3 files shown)


dinosaure:
@hhugo this PR still is actual? Can we review it as it? Thanks for your patch.

hhugo:
> @hhugo this PR still is actual? Can we review it as it? Thanks for your patch.

I've updated the PR and un-drafted it. Ready for review.

dinosaure:
Thanks!

PR #405, opened 2024-05-10
fix extra path separator in mkdir_p

### LINKED ISSUE #404: Extra path separator prepended in mkdir_p
The `mkdir_p` implementation adds an extra `/` at the start of paths.

On linux this causes no issues, but under cygwin the path gets interpreted as a network share (e.g. `//host/share`), and the test run fails with `Unix.EROFS`

bbonf:
Closes #404

### DIFF
--- a/src/alcotest/alcotest.ml
+++ b/src/alcotest/alcotest.ml
@@ -21,7 +21,7 @@ module Unix_platform (M : Alcotest_engine.Monad.S) = struct
             mk path names
       in
       match String.cuts ~empty:true ~sep path with
-      | "" :: xs -> mk sep xs
+      | "" :: xs -> mk "" xs
       (* check for Windows drive letter *)
       | dl :: xs when is_win_drive_letter dl -> mk dl xs
       | xs -> mk "." xs


MisterDA:
@bbonf I can't reproduce the error. How did you encounter it? Could you add a failing test case perhaps?

bbonf:
I'm not sure how to approach adding a test case for this, but here's how I reproduce it:

1. Installed ocaml on cygwin (windows 10 build 19045)
2. Created a new switch with `ocaml-system.4.14.2`
3. `opam install dune alcotest`
4. `dune init project foo` (in `/tmp` but I don't think it matters)
5. Added `(libraries alcotest)` to `test/dune`
6. Pasted this into `test/test_foo.ml`:

```ocaml
open Alcotest

let test () =
  (check @@ int) "just a test"
    0
    1

let () =
  run "Bar" [
    "test", [
      test_case "something" `Quick test
    ]
  ]
```

7. Run `dune test`

MisterDA:
> Installed ocaml on cygwin (windows 10 build 19045)

Do you mean that you're using the Cygwin port of OCaml (or the OCaml package from the Cygwin package archive)?
Or that you're using a Cygwin shell (and have the MinGW-w64 or MSVC ports of OCaml)?

bbonf:
Installed the OCaml package from the Cygwin package archive.

(I think I tried and failed to use the mingw port but that's another story)

PR #412, opened 2024-09-05
testable for seq and contramap (for testable)

xvw:
This PR adds:
- `seq` (to produces `'a Seq.t testable`) Because the `Seq.t` type has become very common in the standard library
- `contramap` which allows you to go from an `'a testable` to a `'b testable` giving a function from `b -> 'a` (because a testable is a contravariant functor). I have often found the function useful for implementing testables _quickly_ when there is a trivial transformation.

### DIFF
--- a/src/alcotest-engine/test.ml
+++ b/src/alcotest-engine/test.ml
@@ -42,7 +42,7 @@ let testable (type a) (pp : a Fmt.t) (equal : a -> a -> bool) : a testable =
   end in
   (module M)
 
-let contramap f t =
+let map f t =
   let pp ppf = (Fmt.using f (pp t)) ppf and equal a b = equal t (f a) (f b) in
   testable pp equal
 
--- a/src/alcotest-engine/test.ml
+++ b/src/alcotest-engine/test.ml
@@ -42,6 +42,11 @@ let testable (type a) (pp : a Fmt.t) (equal : a -> a -> bool) : a testable =
   end in
   (module M)
 
+let contramap f t =
+  let pp ppf = (Fmt.using f (pp t)) ppf
+  and equal a b = (equal t (f a) (f b)) in
+  testable pp equal
+
 let int32 = testable Fmt.int32 ( = )
 let int64 = testable Fmt.int64 ( = )
 let int = testable Fmt.int ( = )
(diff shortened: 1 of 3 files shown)


src/alcotest-engine/test.mli
```diff
@@ -107,6 +110,10 @@ val neg : 'a testable -> 'a testable
 (** [neg t] is [t]'s negation: it is [true] when [t] is [false] and it is
     [false] when [t] is [true]. *)
 
+val contramap : ('b -> 'a) -> 'a testable -> 'b testable
```
samoht, line 113: Do you mind naming this `map` -- even if it doesn't follow the theory, that's what more users would expect :p

samoht:
LGTM apart questions on names (the most difficult questions, as usual)

src/alcotest-engine/test.mli
xvw: What do you think about keeping the `Fmt` convention by calling it `using`? (Because I find that calling it `map` makes reading its signature ... complicated).

xvw: Hm, btw you are right, `map` works well, fixed on https://github.com/mirage/alcotest/pull/412/commits/415abb42b1a1a07bce3a6230a19d1bb4ab513a40

samoht: `using` could work too - no strong opinion between `map` and `using`

xvw: I go for `map` :)

src/alcotest-engine/test.ml
```diff
@@ -78,6 +82,15 @@ let list e =
   in
   testable (Fmt.Dump.list (pp e)) eq
```
MisterDA, line 84:
```suggestion
module Seq = struct
  include Seq
  
  (* Seq.uncons was introduced in OCaml 4.14 *)
  let uncons xs =
    match xs() with
    | Cons (x, xs) ->
        Some (x, xs)
    | Nil ->
        None
end
```

MisterDA (changes requested):
`Seq.uncons` was introduced in OCaml 4.14. We may want to keep supporting OCaml 4.08 for a while still.

src/alcotest-engine/test.ml
xvw: Thanks for the review, in https://github.com/mirage/alcotest/pull/412/commits/47e5f3739a18372b818278ff415297d3dc76ee6f I rewrite `eq` without using `uncons`. Ok for you @MisterDA ?

MisterDA: yep

MisterDA (approved):

PR #413, opened 2024-11-19
no longer rely on joo_global_object

hhugo:
It's deprecated and will result in a warning with the next release of jsoo


MisterDA (approved):

MisterDA:
Does this need a lower or upper bound on jsoo?

hhugo:
You need js_of_ocaml-compiler >= 4.0.0 but you already have a lower bound set to 5.8

MisterDA (approved):

PR #416, opened 2024-12-23
engine: expose skip exception

### LINKED ISSUE #415: Expose Skip exception
For ocaml-junit support of alcotest it is necessary to catch the `Alcotest_engine.V1.Core.Skip` exception to detect tests which are skipped.

https://github.com/Khady/ocaml-junit/pull/10/files#diff-8485feec8501e5d458218068a62dcc2de7a7ea65f0f718f7798bcf097d1ac657R37-R42

So far this exception is not officially exposed, even though it can be accessed through `Alcotest_engine__core`.

Khady:
Fix #415

### DIFF
--- a/src/alcotest-engine/test.ml
+++ b/src/alcotest-engine/test.ml
@@ -268,4 +268,4 @@ let match_raises ?here ?pos msg exnp f =
             Fmt.pf ppf "%t%a %s: got %a." (pp_location ?here ?pos) Pp.tag `Fail
               msg Fmt.exn e)
 
-let skip () = raise Core.Skip
+let skip () = raise Core.V1.Skip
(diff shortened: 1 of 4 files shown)


src/alcotest-engine/test.ml
```diff
@@ -268,4 +268,4 @@ let match_raises ?here ?pos msg exnp f =
             Fmt.pf ppf "%t%a %s: got %a." (pp_location ?here ?pos) Pp.tag `Fail
               msg Fmt.exn e)
 
-let skip () = raise Core.Skip
+let skip () = raise Core.V1.Skip
```
Khady, line 271: I wasn't too sure if test.ml was suppose to depend on V1 or not.

MisterDA (approved):
LGTM

src/alcotest-engine/test.ml
samoht: yea that's a bit odd and seems unrelated with the rest of the changes?

samoht:
The lower bound issue should be fixed by https://github.com/ocaml/opam-repository/pull/27616

PR #424, opened 2025-09-30
Update tests to cmdliner 2.0.0

MisterDA:


dinosaure:
This test should probably be removed since it concerns `cmdliner`, and @dbuenzli [has explicitly stated that the output should not be trusted and used as a test](https://discuss.ocaml.org/t/reminder-opam-repository-package-tests-do-not-snapshot-test-cmdliners-end-user-outputs/17331/3).

MisterDA:
I'm fine with removing the test, too.

PR #433, opened 2026-02-20
Base implementation of catching exceptions made in C stubs

### LINKED ISSUE #430: Alcotest silently fails when aborted from C stub
This example doesn't show if test failed at all.

```c
CAMLprim value caml_crashing_function()
{
  CAMLparam0();
  int* p = NULL;
  *p = 0;
  CAMLreturn(Val_unit);
}
```
```ocaml
(* link with cstub*)
external crashing_function: unit -> unit = "caml_crashing_function"

let () = 
  run "Tests" ["Crash test", [test_case "call crashing function" `Quick (function _ -> crashing_function ())]]
```

There's no output and if the test has crashed somewhere in the middle of test, remaining suite isn't executed and it won't show any error.

Environment: win11 x86-64 using msys2-mingw64

SturdyPose:
The goal of this PR is to fix https://github.com/mirage/alcotest/issues/430

- [X] Add base implementation for catching segmentation faults with try catch block
- [X] Implementation for windows stacktraces
- [x] Implementation for linux stacktraces
- [x] Credit author for collecting windows stack trace in the comments - https://smhk.net/note/2025/03/c-stack-trace-in-windows/
- [x] Update stubs.expected file
- [x] Export different exception structs like SegFault
- [x] Make working JS backend build

Open to any feedback. :)


src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+#include <stdlib.h>
+
+const size_t buffer_size = 1024 * 10;
+typedef struct {
+  char *buffer;
+  size_t capacity;
+  size_t offset;
+} StackTraceBuffer;
+
+bool finit_stack_trace_buffer(StackTraceBuffer* pStackTraceBuffer, size_t size)
```
MisterDA:
```suggestion
static bool finit_stack_trace_buffer(StackTraceBuffer* pStackTraceBuffer, size_t size)
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+{
+  char *error_buffer = (char *)malloc(size);
+  if (error_buffer == NULL) {
+    return false;
+  }
+  pStackTraceBuffer->capacity = size;
+  pStackTraceBuffer->offset = 0;
+  pStackTraceBuffer->buffer = error_buffer;
+  pStackTraceBuffer->buffer[0] = '0';
```
MisterDA:
```suggestion
  pStackTraceBuffer->buffer[0] = '\0';
```
?

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+  return true;
+}
+
+void append_to_buffer(StackTraceBuffer *sb, const char *format, ...) {
```
MisterDA:
```suggestion
static void append_to_buffer(StackTraceBuffer *sb, const char *format, ...) {
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+
+static const char *CAML_ERROR_ID = "segfault exception";
+
+#if defined(_WIN32) || defined(_WIN64)
+#define PLATFORM_WINDOWS
+#include <excpt.h>
+#include <windows.h>
+#include <dbghelp.h>
+
+void create_stacktrace(StackTraceBuffer* pStackTraceBuffer)
```
MisterDA:
```suggestion
static void create_stacktrace(StackTraceBuffer* pStackTraceBuffer)
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+      // Truncated or filled exactly; ensure null termination at the end
+      sb->offset = sb->capacity - 1;
+      sb->buffer[sb->offset] = '\0';
+    }
+  }
+}
+
+static const char *CAML_ERROR_ID = "segfault exception";
+
+#if defined(_WIN32) || defined(_WIN64)
```
MisterDA:
```suggestion
#if defined(_WIN32)
```

`_WIN32` is also defined under Windows 64 bits.

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+    HANDLE       thread  = GetCurrentThread();
+    CONTEXT      context;
+    STACKFRAME64 stack;
+    DWORD        machine_type;
+
+    RtlCaptureContext(&context);
+
+    ZeroMemory(&stack, sizeof(STACKFRAME64));
+
+#ifdef _M_IX86
```
MisterDA:
```suggestion
#if defined(_M_IX86) || defined(__i386__)
```
The first value is only defined under MSVC, the second with GCC.

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+    machine_type           = IMAGE_FILE_MACHINE_I386;
+    stack.AddrPC.Offset    = context.Eip;
+    stack.AddrFrame.Offset = context.Ebp;
+    stack.AddrStack.Offset = context.Esp;
+#elif _M_X64
```
MisterDA:
```suggestion
#elif defined(_M_X64) || defined(__x86_64__)
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+    machine_type           = IMAGE_FILE_MACHINE_AMD64;
+    stack.AddrPC.Offset    = context.Rip;
+    stack.AddrFrame.Offset = context.Rsp;
+    stack.AddrStack.Offset = context.Rsp;
+#elif _M_ARM64
```
MisterDA:
```suggestion
#elif defined(_M_ARM64) || defined(__aarch64__)
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+        SymFunctionTableAccess64,
+        SymGetModuleBase64,
+        NULL)) {
+        if (stack.AddrPC.Offset == 0)
+            break;
+
+        DWORD64 symbol_addr  = stack.AddrPC.Offset;
+        DWORD64 displacement = 0;
+        char symbol_buffer[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)] = {0};
+        SYMBOL_INFO *symbol  = (SYMBOL_INFO *)symbol_buffer;
```
MisterDA:
```c
        char symbol_buffer[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)] = {0};
        SYMBOL_INFO *symbol  = (SYMBOL_INFO *)symbol_buffer;
```
Even if this works now this pattern is generally bad because `SYMBOL *` has a stricter alignment than `char`, so you may end up with a [misaligned pointer](https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Aliasing-Alignment.html). Either use a union:
```suggestion
        union {
          char buf[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)];
          SYMBOL_INFO *symbol;
        } u;
        memset(&u, 0, sizeof(u));
        u.symbol = (SYMBOL_INFO *)symbol_buffer;
```
or `alignas`:
```suggestion
        _Alignas(SYMBOL_INFO *) char symbol_buffer[sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(TCHAR)] = {0};
        SYMBOL_INFO *symbol  = (SYMBOL_INFO *)symbol_buffer;
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+    return EXCEPTION_CONTINUE_SEARCH; 
+}
+#else
+#define PLATFORM_UNIX
+#include <signal.h>
+#include <execinfo.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+void unix_signal_handler(int sig, siginfo_t *si, void *unused) {
```
MisterDA:
```suggestion
static void unix_signal_handler(int sig, siginfo_t *si, void *unused) {
```

test/e2e/alcotest/stubs/nullexceptionstub.c
```diff
@@ -0,0 +1,17 @@
...
+#include <caml/callback.h>
+#include <caml/config.h>
+#include <caml/custom.h>
+#include <caml/fail.h>
+#include <caml/intext.h>
+#include <caml/memory.h>
+#include <caml/misc.h>
+#include <caml/mlvalues.h>
+
+value caml_segfault_call()
```
MisterDA:
```suggestion
value caml_segfault_call(void)
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,236 @@
...
+  free(pSymbols);
+
+  caml_raise_with_string(*caml_named_value(CAML_ERROR_ID),
+                         stack_trace_buffer.buffer);
+  exit(WEXITED);
+}
+#endif
+
+
+CAMLprim value caml_setup_stub_exception_handler()
```
MisterDA:
```suggestion
CAMLprim value caml_setup_stub_exception_handler(void) 
```

MisterDA (approved):
There's a bit of mixed indentation in the C file, please reformat the Windows code to be consistent with the rest.

src/alcotest-engine/stack_error_reporter.c
SturdyPose: Oh crap. Missed this! Thanks!

src/alcotest-engine/stack_error_reporter.c
SturdyPose: Didn't know that! Thank you! :)

SturdyPose:
Applied clang-format on source

SturdyPose:
Seems like windows containers are broken because they don't have npm and I'm unsure why mac is not catching the exception... 🤔

src/alcotest-engine/dune
MisterDA, line 1: What is happening here? Why has the whole file changed?

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+      sb->offset = sb->capacity - 1;
+      sb->buffer[sb->offset] = '\0';
+    }
+  }
+}
+
+static const char *CAML_ERROR_ID = "segfault exception";
+
+#if defined(_WIN32)
+#define PLATFORM_WINDOWS
```
MisterDA:
```suggestion
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+    // Print with better alignment using format specifiers
+    append_to_buffer(pStackTraceBuffer, "    %-40.40s 0x%016llX %s\n",
+                     function_name, symbol_addr, line_info);
+    append_to_buffer(pStackTraceBuffer, "\0");
+  }
+
+  SymCleanup(process);
+}
+
+LONG WINAPI windows_exception_handler(PEXCEPTION_POINTERS pExceptionInfo) {
```
MisterDA:
```suggestion
static LONG WINAPI windows_exception_handler(EXCEPTION_POINTERS *pExceptionInfo) {
```

Prefer not to use the pointer typedefs but rather a pointer to the underlying type. It allows making the pointer `const`.

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+    free(stack_trace_buffer.buffer);
+    ExitProcess(STATUS_ACCESS_VIOLATION);
+  }
+  default:
+    break;
+  }
+  return EXCEPTION_CONTINUE_SEARCH;
+}
+#else
+#define PLATFORM_UNIX
```
MisterDA:
```suggestion
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+                         stack_trace_buffer.buffer);
+  exit(WEXITED);
+}
+#endif
+
+CAMLprim value caml_setup_stub_exception_handler(void) {
+  CAMLparam0();
+#ifdef PLATFORM_WINDOWS
+  AddVectoredExceptionHandler(1, windows_exception_handler);
+#elif defined(PLATFORM_UNIX)
```
MisterDA:
```suggestion
#ifdef _WIN32
  AddVectoredExceptionHandler(1, windows_exception_handler);
#else
```

test/e2e/alcotest/stubs/nullexceptionstub.c
```diff
@@ -0,0 +1,17 @@
+#include <caml/alloc.h>
+#include <caml/callback.h>
+#include <caml/config.h>
+#include <caml/custom.h>
+#include <caml/fail.h>
+#include <caml/intext.h>
+#include <caml/memory.h>
+#include <caml/misc.h>
+#include <caml/mlvalues.h>
```
MisterDA:
```suggestion
#include <caml/memory.h>
#include <caml/mlvalues.h>
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
+// caml headers
+#include <caml/alloc.h>
+#include <caml/callback.h>
+#include <caml/config.h>
+#include <caml/custom.h>
+#include <caml/fail.h>
+#include <caml/intext.h>
+#include <caml/memory.h>
+#include <caml/misc.h>
+#include <caml/mlvalues.h>
```
MisterDA, line 6:
```suggestion
#include <caml/alloc.h>
#include <caml/callback.h>
#include <caml/custom.h>
#include <caml/fail.h>
#include <caml/intext.h>
#include <caml/memory.h>
#include <caml/misc.h>
#include <caml/mlvalues.h>
```

Not sure that you need all these? maybe `misc.h` and `intext.h` are superfluous?

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+#include <caml/intext.h>
+#include <caml/memory.h>
+#include <caml/misc.h>
+#include <caml/mlvalues.h>
+
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
```
MisterDA, line 10:
```suggestion
#include <stdalign.h>

```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+                   "-------", "----");
+
+  while (StackWalk64(machine_type, process, thread, &stack, &context, NULL,
+                     SymFunctionTableAccess64, SymGetModuleBase64, NULL)) {
+    if (stack.AddrPC.Offset == 0)
+      break;
+
+    DWORD64 symbol_addr = stack.AddrPC.Offset;
+    DWORD64 displacement = 0;
+    _Alignas(SYMBOL_INFO *)
```
MisterDA:
```suggestion
    alignas(SYMBOL_INFO *)
```

src/alcotest-engine/stack_error_reporter.c
```diff
@@ -0,0 +1,230 @@
...
+  void *trace[STACK_TRACE_LENGTH];
+  size_t trace_size = backtrace(trace, STACK_TRACE_LENGTH);
+
+  if (trace_size == 0) {
+    caml_failwith("Couldn't get backtrace");
+    return;
+  }
+
+  append_to_buffer(&stack_trace_buffer,
+                   "Caught Violation access, here's stack trace:\n");
```
MisterDA:
```suggestion
  append_to_buffer(&stack_trace_buffer, "Access violation caught, stacktrace:\n");
```

MisterDA (changes requested):

SturdyPose:
Yeah, I'll need to rethink the implementation of reporting it back to ocaml -> might not be possible at all. There can be stack corruption and that would lead to undeterministic results. Do you think that reporting it to stderr would be good enough?

MisterDA:
For the unix part maybe you need a signal handler on the ocaml side instead?

SturdyPose:
Well from what I gathered online is that signals can be interpreted differently for each unix like system. For example apple platform is handling signals differently than linux systems, and even linux distros themselves can handle those differently, so I reckon that it won't make much difference adding signal handler on ocaml side. There's also problem when you throw a signal like SEGFAULT, you can't reliably catch it and especially returning back to ocaml's managed memory. I think that only course of action is to just print into stderror.

 Such a headache. :D

SturdyPose:
I don't like to keep things open ended, @MisterDA I'm thinking about solving this with just reporting failure point and letting rest of the tests fail as well. How does that sound? I can update the implementation to match it.

MisterDA:
> I don't like to keep things open ended, @MisterDA I'm thinking about solving this with just reporting failure point and letting rest of the tests fail as well. How does that sound? I can update the implementation to match it.

Sounds good!

PR #439, opened 2026-05-11
Alcotest.run: ensure and_exit and record_backtrace can be overriden f…

### LINKED ISSUE #438: and_exit cannot be overriden when using test subcommand
When using `Alcotest.run ~and_exit:false` the test will still call `exit 0` when the `test` subcommand is used (all works fine when it runs automatically with the default command).
This is a problem because cleanup handlers registered with `Fun.protect` won't run.

To reproduce:

[alcobug.zip](https://github.com/user-attachments/files/27590363/alcobug.zip)

```
dune runtest
File "test/dune", line 2, characters 7-19:
2 |  (name test_alcobug)
           ^^^^^^^^^^^^
Testing `bug'.
This run has ID `NE0Z2SZ9'.

  [OK]          bug          0   Bug.

Test Successful in 0.000s. 1 test run.

[36185]: Code before Fun.protect

[36185]: at_exit

[36185]: exit called, but finally didn't run yet!
```

This is probably due to this line:
```
alcotest.1.9.1/src/alcotest-engine/cli.ml
133:        Config.User.term ~stdout ~stderr ~and_exit:true ~record_backtrace:true
```

edwintorok:
…or the test sub-command too

`default_cmd` handles this correctly, but `test_cmd` always set `and_exit` and `record_backtrace` to `true`.

Due to `Config.User.(cli_config || config)` this meant that `and_exit` couldn't be set to `false` by the caller.
The main program may want to set `and_exit` to false if it has some wrappers using `Fun.protect` for example, with `and_exit` set to true these wouldn't run.

For backwards compatibility with old versions of alcotest one workaround is to use `at_exit` instead of `Fun.protect` to register cleanup handlers.

Fixes https://github.com/mirage/alcotest/issues/438


edwintorok:
The CI needs some maintenance due to the removal of <4.11 from opam-repository.

